PR #202, opened 2014-11-13
init: uid/gid adaption

fetzerch:
Thanks for this dockerfile :)

I'm running a small gitlab docker-environment together with a few other non dockerized services on a server. The repositories are located in git's home under /home/git/data on the host. This change allows me to make sure that the user git on the host owns all the files in /home/git/data which is mounted into the container.
Otherwise they would be owned by 1000:1000 which belongs to a completely different user. A nice (but unimportant) side effect is that a ps on the host shows that related daemons are running as user git (instead of the user associated with 1000). More details can be found in the commit message.

Since this change might be interesting for other users as well I've created this PR.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 **latest**
 - added `SSL_VERIFY_CLIENT` configuration option
 - added `NGINX_WORKERS` configuration option
+- added `USERMAP_UID` and `USERMAP_GID` configuration option
 
 **7.4.3**
 - gitlab: upgrade to CE v.7.4.3


jks20:
I'm having a small issue with the USERMAP_GID. If the group already exists inside the docker, the user isn't added to the group and initialization of the docker fails.

sameersbn:
@jks20 yep. such cases are not handled.

jks20:
@sameersbn Are there any thoughts of adding this ability or is it a security risk?

sameersbn:
@jks20 could be added. There could technically be a security risk, but I think its highly unlikely.

jks20:
Awesome! And thanks a lot for this dockerfile. Setting it up has been a breeze!

Issue #208, opened 2014-11-17
Sendmail


msteinert:
This is more of a question than an issue. I found your instructions for using config overrides and that part worked great. Next I added sendmail to my image and started it in the init script (msteinert/docker-gitlab@edf390b113530ba4d1adedfe3c5e379931158bbf). Is there a better way to do this?

I suspect this change is too specialized to add to your base configuration. Perhaps there might be some way to add a sendmail configuration variable and use that to decide whether or not to start sendmail?

Cheers!

sameersbn:
@msteinert should we close this issue?

msteinert:
Sorry, I forgot about this. The solution suggested by @skython worked perfectly.

PR #211, opened 2014-11-20
Add configuration for sendmail

msteinert:
If `SENDMAIL_ENABLED` is set to `true` then postfix will be started and Gitlab will be configured to use sendmail. If both `SENDMAIL_ENABLED` and `SMTP_ENABLED` are `true` then the sendmail feature is disabled. The configuration for postfix is simple; it is setup as a relay.

Postfix forks into the background and as a result cannot be started under supervisord. See the following link for more background information:

http://goo.gl/F4neAQ

Closes sameersbn/docker-gitlab#208

I opted to use postfix over sendmail because postfix is the default MTA on Ubuntu and is significantly easier to configure.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apt-key adv --keyserver keyserver.ubuntu.com --recv E1DF1F24 \
  && wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
  && echo 'deb http://apt.postgresql.org/pub/repos/apt/ trusty-pgdg main' > /etc/apt/sources.list.d/pgdg.list \
  && apt-get update \
- && apt-get install -y supervisor logrotate locales \
+ && apt-get install -y supervisor logrotate locales postfix \
       nginx openssh-server mysql-client postgresql-client redis-tools \
       git-core ruby2.1 python2.7 python-docutils \
       libmysqlclient18 libpq5 zlib1g libyaml-0-2 libssl1.0.0 \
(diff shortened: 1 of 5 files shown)


skython:
I think sendmail just doesn't belong into the GitLab container and using postfix volumes with a socket sounds a bit messy.

Why not stick with SMTP to decouple both? 

Its pretty easy actually:

**Start postfix:**
```
docker run -d --name gitlab_postfix \
 -e "maildomain=YOURDOMAIN.ORG" \
 -e "smtp_user=gitlab:YOURPASSWORD" \
 catatnight/postfix
```

**Start GitLab linked to the postfix container:**
```
 ...
 --link gitlab_postfix:postfix \
 -e "SMTP_DOMAIN=YOURDOMAIN.ORG" \
 -e "SMTP_USER=gitlab" \
 -e "SMTP_PASS=YOURPASSWORD" \
 -e "SMTP_HOST=postfix" \
 -e "SMTP_PORT=25" \
 -e "SMTP_STARTSSL=false" \
 -e "SMTP_OPENSSL_VERIFY_MODE=none" \
 ...
```

Its internal, so i dropped SSL in this configuration.

msteinert:
That's a good idea, I'll have to try it out.

I agree, including Postfix in the Gitlab image is probably ill-advised.

kain88-de:
@skython did this setup work for you? I have created the docker containers locally like you described but I never get an email from this setup.

the `production.log` from gitlab says that it has send all emails. But when I check the `mail.log` in the postfix container I see no attempt to send a mail at all. I see that gitlab attempts to send email when I use `smtpd_relay_restrictions` in `/etc/postfix/main.cf`. But all those attempts say that the SASL Authentication failed. 

Was there maybe anything missing in your setup?

PR #240, opened 2015-01-22
Allow enabling AWS backups with environment variables

andycmaj:


sameersbn:
@andycmaj have you tested the backup operation when user does not configure remote backup.

PR #269, opened 2015-02-28
possibility to disable proxy buffering in nginx or in a frontend nginx

ulrichSchreiner:
Hi,

here are the changes i'm thinking of. feel free to discard this PR. i think all three settings have their defaults.


sameersbn:
Hi. Can you rebase the PR. I will merge it first thing in the morning before making the 7.9.3 release. Let me know. Thanks.

sameersbn:
@ulrichSchreiner Do you thing just setting `proxy_buffering off;` by default should be sufficient given that the application is running inside a docker container. Maybe we also want to default `add_header X-Accel-Buffering no;`. The `proxy_max_temp_file_size` option I think is overkill.

So basically we can have

```nginx
add_header X-Accel-Buffering {{NGINX_ACCEL_BUFFERING}};
proxy_buffering          {{NGINX_PROXY_BUFFERING}};
```

With both these disabled by default so that the defaults work as is for most users. Let me know what you think. If you agree, I will merge the PR and make the required changes.

Thanks.

ulrichSchreiner:
hi,

If you disable both settings as default most users won't need the `max_temp_file_size`. so both settings should be enough, you're right.

sameersbn:
@ulrichSchreiner ok.. I will change defaults and get rid of the `max_temp_file_size` configuration

PR #273, opened 2015-03-12
Allow setting of the root password

russell:
Hey,

This patch exposes an environment variable that can be used to configure a default root password.

The only documentation I have seen for this is here https://github.com/gitlabhq/gitlabhq/blob/master/doc/install/installation.md#initialize-database-and-activate-advanced-features

Cheers,
Russell

### DIFF
--- a/assets/init
+++ b/assets/init
@@ -731,7 +731,7 @@ appStart () {
   esac
   if [ -z "${COUNT}" -o ${COUNT} -eq 0 ]; then
     echo "Setting up GitLab for firstrun. Please be patient, this could take a while..."
-    sudo -u git -H force=yes bundle exec rake gitlab:setup RAILS_ENV=production >/dev/null
+    sudo -u git -H force=yes bundle exec rake gitlab:setup RAILS_ENV=production ${GITLAB_ROOT_PASSWORD:+GITLAB_ROOT_PASSWORD=$GITLAB_ROOT_PASSWORD} >/dev/null
   fi
 
   # migrate database and compile the assets if the gitlab version or relative_url has changed.
--- a/README.md
+++ b/README.md
@@ -714,6 +714,7 @@ Below is the complete list of available options that can be used to customize yo
 - **GITLAB_HOST**: The hostname of the GitLab server. Defaults to `localhost`
 - **GITLAB_PORT**: The port of the GitLab server. Defaults to `80` for plain http and `443` when https is enabled.
 - **GITLAB_TIMEZONE**: Configure the timezone for the gitlab application. This configuration does not effect cron jobs. Defaults to `UTC`.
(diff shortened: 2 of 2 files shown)


sameersbn:
@russell Thanks for the PR. But do you think this option is really required? Gitlab is slowly moving all configurable parameters to the webui, which is great because we can remove a lot of configuration options from this image leaving the user to primarily configure only the database, redis, url and ssl configuration.

Besides this PR has the side-effect that if the `GITLAB_ROOT_PASSWORD` variable is not specified, the default password is set. But gitlab will not ask you to change the default password on first login as it normally would.

russell:
So we are hoping to use the web API to do some post configuration of the docker.  But we don't want our users to be able to change the root password before we have set everything up.  E.g. connect up gitlab-ci and add a service account.

You are right, i didn't intend on having that side effect.  I actually can simplify things by just directly passing the env variable without setting a default.  That will keep the default behaviour if the variable isn't specified.

Does this seem reasonable or do you think there is a better way to accomplish this same thing?  I had considered inheriting your image and adding more db init scripts.  But using web API seems like it will be easier to maintain.

sameersbn:
@russell sounds good to me. please update the PR after fixing the side-effect issue. thanks.

russell:
yeah it should be.  i pushed microseconds before commenting.

assets/init
```diff
@@ -731,7 +731,7 @@ appStart () {
   esac
   if [ -z "${COUNT}" -o ${COUNT} -eq 0 ]; then
     echo "Setting up GitLab for firstrun. Please be patient, this could take a while..."
-    sudo -u git -H force=yes bundle exec rake gitlab:setup RAILS_ENV=production >/dev/null
+    sudo -u git -H force=yes bundle exec rake gitlab:setup RAILS_ENV=production GITLAB_ROOT_PASSWORD=${GITLAB_ROOT_PASSWORD} >/dev/null
```
sameersbn: @russell I would suggest replacing `GITLAB_ROOT_PASSWORD=${GITLAB_ROOT_PASSWORD}` with `${GITLAB_ROOT_PASSWORD:+GITLAB_ROOT_PASSWORD=$GITLAB_ROOT_PASSWORD}` to avoid any breakages in the future.

`${GITLAB_ROOT_PASSWORD:+GITLAB_ROOT_PASSWORD=$GITLAB_ROOT_PASSWORD}` will make sure the `GITLAB_ROOT_PASSWORD` parameter is added to the command line only if `GITLAB_ROOT_PASSWORD` is defined.

sameersbn: @russell also, please rebase the PR while your at it. thanks.

russell: Thanks for the feedback.  I didn't know that bash form!

PR #278, opened 2015-03-21
Updated README in order to fix libdevmapper.so issues

m451:
Added libdevmapper.so mapping which seems to fix the error "docker: error while loading shared libraries: libdevmapper.so.1.02: cannot open shared object file: No such file or directory" which appears in coreos and other distributions as of the latest docker release.

Feel free to test with other distributions. The error seems to be common.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,7 @@ docker run --name='gitlab' -it --rm \
 -p 10022:22 -p 10080:80 \
 -v /var/run/docker.sock:/run/docker.sock \
 -v $(which docker):/bin/docker \
+-v /usr/lib/libdevmapper.so.1.02:/usr/lib/libdevmapper.so.1.02 \
 sameersbn/gitlab:7.8.4
 ```
 


sameersbn:
@m451 This won't work on systems that don't have the `/usr/lib/libdevmapper.so.1.02` file. `ubuntu:14.04` in my case.

m451:
True. Seems like it needs more investigation. Multiple users report that error.

sameersbn:
@m451 solution is to just use `fig` or `docker-compose` instead of mounting the docker binary and socket. I will be nuking that option soon.

sameersbn:
@m451 If you are facing this issue can you try the following command

```bash
docker run --name='gitlab' -it --rm \
-e 'GITLAB_PORT=10080' -e 'GITLAB_SSH_PORT=10022' \
-p 10022:22 -p 10080:80 \
-v /var/run/docker.sock:/run/docker.sock \
-v $(which docker):/bin/docker \
-v $(ldconfig -p | grep libdevmapper | awk '{print $4}'):/usr/lib/libdevmapper.so.1.02 \
sameersbn/gitlab:7.8.4
```

I have not faced this issue so cannot test. Please let me know.

sameersbn:
volume mounting the docker binary and socket in the quick start guide has been replaced with manually start of postgresql and redis images or using docker-compose to orchestrate the startup.

PR #287, opened 2015-04-12
README: add a missing configuration OAUTH_ENABLED

chengweiv5:
Signed-off-by: Chengwei Yang <yangchengwei@qiyi.com>


sameersbn:
@ chengweiv5 actually this is not missing configuration. techincally `OAUTH_ENABLED` should be viewed an an internal script available which is automatically enabled only when one or more OAUTH configurations are specified. The `init` script makes sure that all required parameters have been specified before setting omniauth enabled.

sameersbn:
@chengweiv5 thanks for the contribution. Changes have been merged

PR #288, opened 2015-04-14
Use long GPG key ids for security

lazyfrosch:
This avoids possible flaws against duplicate or manipulated keys.

The Postgres keys should be fine, because it gets downloaded via https.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
 FROM sameersbn/ubuntu:14.04.20150323
 MAINTAINER sameer@damagehead.com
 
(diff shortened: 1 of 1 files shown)


sameersbn:
@lazyfrosch although I have merged this PR. Does specifying the full key really make a difference security wise? I seriously think it will not make a difference whatsoever.

lazyfrosch:
@sameersbn well, it will when someone tries to do a MITM attack on the keyserver, HTTPs can be verified against the ca-certificates bundle.

Short GPG ids can be faked easily.

PR #296, opened 2015-04-19
Create app:init to separate initialization from starting

aforward:
This feature allows an install script to initialize much of the
gitlab requirements without actually starting it.  It is useful
for deployment scripts so that the first start up does not have to
do this work itself.  app:start calls app:init, so no change
in functionality for app:start.

### DIFF
--- a/assets/init
+++ b/assets/init
@@ -721,7 +721,7 @@ chmod 700 ${GITLAB_DATA_DIR}/.ssh
 chmod 600 ${GITLAB_DATA_DIR}/.ssh/authorized_keys
 chown -R git:git ${GITLAB_DATA_DIR}/.ssh
 
-appStart () {
+appInit () {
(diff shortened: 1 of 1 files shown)


assets/init
```diff
@@ -721,7 +721,7 @@ chmod 700 ${GITLAB_DATA_DIR}/.ssh
 chmod 600 ${GITLAB_DATA_DIR}/.ssh/authorized_keys
 chown -R git:git ${GITLAB_DATA_DIR}/.ssh
 
-appStart () {
+appInit () {
```
sameersbn: @aforward I think you should modify the `appInit` function to `return 1` on failures. i.e. replace the `exit 1` with `return 1` and end the function with `return 0`

sameersbn: @aforward ok maybe not. un-read my previous comment

PR #303, opened 2015-04-27
Add GIT_TIMEOUT configuration parameter

Leibniz137:
Defaults to 10s (the current hardcoded value)

Testing Done:
 - Confirmed parameter information was correctly rendered in README.md
 - Set parameter to a non-default value, confirmed that this value was set in gitlab.yml
 - Didn't set parameter, confirmed value was set to default 30s in gitlab.yml


Leibniz137:
Correction: 10s, not 30s (copy/paste error from previous merge request)

sameersbn:
@Leibniz137 can you rename the parameter to `GITLAB_TIMEOUT` so that the naming is homogenous with the other parameters. Thanks.

PR #268, opened 2015-02-27
parameterize AWS backup config

andycmaj:
tested without specifying AWS provider and config... looks good.

```bash
⮀ docker run --name=gitlabbak -it --rm --link dockergitlab_redis_1:redisio --link dockergitlab_postgresql_1:postgresql sameersbn/gitlab:7.8.1 app:rake gitlab:backup:create
Creating SSH2 RSA key; this may take some time ...
Creating SSH2 DSA key; this may take some time ...
Creating SSH2 ECDSA key; this may take some time ...
Creating SSH2 ED25519 key; this may take some time ...
invoke-rc.d: policy-rc.d denied execution of restart.
Running gitlab rake task...
Dumping database ...
Dumping PostgreSQL database gitlabhq_production ... [DONE]
done
Dumping repositories ...
done
Dumping uploads ...
done
Creating backup archive: 1425080961_gitlab_backup.tar ... done
Uploading backup archive to remote storage  ... skipped
Deleting tmp directories ... done
Deleting old backups ... skipping
```

PR #304, opened 2015-04-29
add AWS_BACKUPS to configure automatic backups to S3

colinbrownec:
Enables use of GitLab integration with Fog to automatically push backups to an S3 bucket.

I noticed there is another [open pull request](https://github.com/sameersbn/docker-gitlab/pull/268) doing this same feature, this setup is much simpler.

I just realized now that Fog supports multiple providers, but I'm not sure how to support complete configuration through environment variables without writing a lot more code.

I've tested this on a production git server and it works with both `app:rake` backups as well as automatic backups. It also disables this behavior when `AWS_BACKUPS=false`

### DIFF
--- a/README.md
+++ b/README.md
@@ -788,6 +788,11 @@ Below is the complete list of available options that can be used to customize yo
 - **GOOGLE_ANALYTICS_ID**: Google Analytics ID. No defaults.
 - **PIWIK_URL**: Sets the Piwik URL. No defaults.
 - **PIWIK_SITE_ID**: Sets the Piwik site ID. No defaults.
+- **AWS_BACKUPS**: Enables automatic uploads to an Amazon S3 instance. Defaults to `false`.
+- **AWS_BACKUP_REGION**: AWS region. No defaults.
+- **AWS_BACKUP_ACCESS_KEY_ID**: AWS access key id. No defaults.
+- **AWS_BACKUP_SECRET_ACCESS_KEY**: AWS secret access key. No defaults.
+- **AWS_BACKUP_BUCKET**: AWS bucket for backup uploads. No defaults.
 
 # Maintenance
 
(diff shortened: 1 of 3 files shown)


sameersbn:
@colinbrownec do you think it would be a better idea to automatically set `AWS_BACKUPS` to `true` when the required parameters are configured, so as to avoid broken configurations?

sameersbn:
@colinbrownec basically we check if `AWS_BACKUP_REGION`, `AWS_BACKUP_ACCESS_KEY_ID`, `AWS_BACKUP_SECRET_ACCESS_KEY` and `AWS_BACKUP_BUCKET` are not empty before enabling AWS backups.

colinbrownec:
@sameersbn I agree. But if not all parameters are valid, should it exit? I see two options;
1. print error message, disable aws, continue
2. print error message, exit

I guess I'm more for 2. since if you have some aws parameters, you were probably trying to use aws

sameersbn:
@colinbrownec I think you can keep the `AWS_BACKUPS` option. But if it is not specified, then it should be enabled automatically if all the four parameters are defined. If it has been enabled by the user, but if any of the four parameters are not defined, the we display a message that says something like "Cannot enable AWS backups. Invalid configuration.", disable AWS backups and continue without exiting.

sameersbn:
@colinbrownec somewhat (not entirBut unlike the case of AWS, ely) similar to `SMTP_ENABLED` configuration. `SMTP_ENABLED` is automatically set to true if SMTP_USER is specified. However `SMTP_ENABLED` can be manually enabled by the user without specifying the `SMTP_USER` which is not the case for `AWS_BACKUPS`. 

`AWS_BACKUPS` has to have the four parameters. I agree with your "2. since if you have some aws parameters, you were probably trying to use aws", but it you would have to carefully craft the code to make sure if `AWS_BACKUPS` was enabled automatically or by the user.

sameersbn:
@colinbrownec In case of OAUTH support if the user does not specify all oauth parameters. I will just not enable oauth and continue. Here the difference it that the user can tell that oauth has not been enabled by looking at the login screen. In case of aws backups this is not the case. So your right, if `AWS_BACKUPS` cannot be enabled when the user was trying to enable it, its best to error out.

sameersbn:
@colinbrownec Lets just keep it simple. If you current PR, add a check to make sure if the four parameters are defined. If not, print error message and exit

sameersbn:
@colinbrownec ps. post a comment here after you update the PR so that i am notified. thanks

colinbrownec:
new code aborts if `AWS_BACKUPS=true` but options are incomplete. I also like this way since you can disable AWS without deleting configuration information.

sameersbn:
Hi. I will test and merge on Monday. Traveling at the moment. Thanks.

PR #324, opened 2015-05-19
Add sidekiq queue archive_repo

likol:
download archive does not work, because sidekiq daemon no listen queue archive_repo


sameersbn:
@likol Thanks for this PR. I would have never realized it. :+1:

Jack12816:
Great! - I realized it yesterday and searched for the reason. Thanks. :)

Issue #342, opened 2015-06-11
Could not start container when using timezone with a slash


urquilla:
After configuring the environment variables and starting the containers using docker-compose received this output:

    gitlab_1     | sed: -e expression #1, char 31: unknown option to `s'
    gitlab_gitlab_1 exited with code 1

Which resulted somewhat cryptic, after fooling around found that using a slash in the timezone variable produced an error with the sed regex.

The workaround was using a backwards slash to escape it:

    - GITLAB_TIMEZONE=America\/El_Salvador

sameersbn:
@urquilla you need to specify the constant that maps to your timezone listed here http://api.rubyonrails.org/classes/ActiveSupport/TimeZone.html.

For example, if I want to set the timezone to `Asia/Kolkata`, I need to use `GITLAB_TIMEZONE=Kolkata`. Please check the constants listed in the above link and use the appropriate one.

urquilla:
I see, thank you very much

sameersbn:
The `&` character is now automatically escaped. If you have manually escaped the `&` character, then you need to remove it in versions > 8.1.x.

PR #343, opened 2015-06-11
Escaping timezone environment variable due use of slashes causing sed error

urquilla:
Proposal for issue #342

Allows entering timezone environment variable without thinking about internal parsing in assets/init

### DIFF
--- a/assets/init
+++ b/assets/init
@@ -346,6 +346,8 @@ sudo -u git -H sed 's/{{GITLAB_SSH_HOST}}/'"${GITLAB_SSH_HOST}"'/' -i config/git
 sudo -u git -H sed 's/{{GITLAB_SSH_PORT}}/'"${GITLAB_SSH_PORT}"'/' -i config/gitlab.yml
 
 # configure default timezone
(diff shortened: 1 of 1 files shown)


thomasleveil:
:+1: nice move. Implementation details (the use of Bash and related escaping issues) should not leak up to the config. Why would anyone have to escape stuff in a environment variable value ?

sameersbn:
@thomasleveil the `GITLAB_TIMEZONE` variable uses ruby TimeZone constants http://api.rubyonrails.org/classes/ActiveSupport/TimeZone.html. If you want to specify `Asia/Kolkata` as your timezone then you specify `GITLAB_TIMEZONE=Kolkata` as per the linked page.

I think this PR is not required unless ruby allows it.

thomasleveil:
IMHO, it sucks a _user_ has to follow a web framework documentation addressed to _developpers_ to be able to set a timezone. Anyway see #352 to help future users to figure this out on their own ;)

sameersbn:
@urquilla closing this PR. thanks!

PR #357, opened 2015-06-23
OAuth : Fix config build error

fredericseiler:
Remove the comma to prevent a nasty "Psych::SyntaxError: (<unknown>): did not find expected '-' indicator while parsing a block collection" when using OAuth Gitlab API environment variables.


sameersbn:
@fredericseiler thanks for the fix.

PR #364, opened 2015-07-06
Allow to override robots.txt

outcoldman:
GitLab has empty robots.txt file by default https://github.com/gitlabhq/gitlabhq/blob/master/public/robots.txt
With this change I allow users to override default one which blocks everything and also allowing users to specify their own files.


sameersbn:
@outcoldman This can just as easily be configured by mounting the robot.txt file using `-v /path/to/robots.txt:/home/git/gitlab/public/robots.txt` on the docker command line.

outcoldman:
@sameersbn true, but it is imposssible when you are trying to avoid mounting anything from the host and use only data volume containers. 
How about more general solution, like allowing to specify GITLAB_CUSTOM_INIT_SCRIPT_PATH, which will be empty by default. This script willl allow to invoke any custom script from specified location before gitlab will be started. 
In that case I will be able to create my own shell script plus with this logic of overriding robots.txt.

PR #370, opened 2015-07-09
Add tutum deploy button

davidwindell:


sameersbn:
@davidwindell I would suggest moving the button before the "Table of Contents" title.

PR #387, opened 2015-07-22
Add `GITLAB_REPOS_DIR` configuration variable

ichizok:
I propose to configure a location of repositories by `GITLAB_REPOS_DIR` variable.
I think it is useful, for example, if we want to construct new container with an existing repositories.


sameersbn:
@ichizok can you document this in the README available configuration options section

ichizok:
I updated README.

PR #402, opened 2015-08-06
The customizing setting is restored

3kami3:
The customizing setting is restored, and the default value is matched up-stream.


sameersbn:
thanks for the fix :+1:

PR #403, opened 2015-08-08
Added `SMTP_CA_ENABLED`, `SMTP_CA_PATH` and `SMTP_CA_FILE` configuration options.

vancepym:
Without ca_path or ca_file, custom CA certs are not taken into account during the
SMTP SSL/TLS handshake. Since the SMTP email configuration allows the use of CApath
and CAfile for OpenSSL, it should be allowed to set these parameters.

See also:
https://gitlab.com/gitlab-org/omnibus-gitlab/commit/fa9c1464bc1eb173660edfded1a2f7add7ac24b3

Signed-off-by: Scott Fan <fancp2007@gmail.com>


sameersbn:
thanks for the contribution.

PR #465, opened 2015-10-12
Fixes #462

### LINKED ISSUE #462: Checking uploads directory permissions...
[entrypoint.sh](https://github.com/sameersbn/docker-gitlab/blob/master/entrypoint.sh#L981)
```
chmod -R 0750 ${GITLAB_DATA_DIR}/uploads/
find ${GITLAB_DATA_DIR}/uploads -type f -exec chmod 0644 {} \;
find ${GITLAB_DATA_DIR}/uploads -type d -not -path ${GITLAB_DATA_DIR}/uploads -exec chmod 0755 {} \;
```

why not this:

```
chmod -R a=rX,u=rwX ${GITLAB_DATA_DIR}/uploads/
chmod 0750 ${GITLAB_DATA_DIR}/uploads/
```
?

Allineer:
Fixes #462

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -845,8 +845,8 @@ chown ${GITLAB_USER}:${GITLAB_USER} ${GITLAB_BACKUP_DIR}
 
 # create the uploads directory
 mkdir -p ${GITLAB_DATA_DIR}/uploads/
-chmod 0750 ${GITLAB_DATA_DIR}/uploads/
-chown ${GITLAB_USER}:${GITLAB_USER} ${GITLAB_DATA_DIR}/uploads/
+chmod -R u=rwX,g=rX,o= ${GITLAB_DATA_DIR}/uploads/
+chown -R ${GITLAB_USER}:${GITLAB_USER} ${GITLAB_DATA_DIR}/uploads/
 
 # create the .ssh directory
 mkdir -p ${GITLAB_DATA_DIR}/.ssh/
@@ -978,10 +978,8 @@ appSanitize () {
   chown -R ${GITLAB_USER}:${GITLAB_USER} ${GITLAB_BUILDS_DIR}
 
   echo "Checking uploads directory permissions..."
-  chmod -R 0750 ${GITLAB_DATA_DIR}/uploads/
-  find ${GITLAB_DATA_DIR}/uploads -type f -exec chmod 0644 {} \;
-  find ${GITLAB_DATA_DIR}/uploads -type d -not -path ${GITLAB_DATA_DIR}/uploads -exec chmod 0755 {} \;
-  chown ${GITLAB_USER}:${GITLAB_USER} ${GITLAB_DATA_DIR}/uploads/
+  chmod -R u=rwX,g=rX,o= ${GITLAB_DATA_DIR}/uploads/
+  chown -R ${GITLAB_USER}:${GITLAB_USER} ${GITLAB_DATA_DIR}/uploads/
 
   echo "Checking tmp directory permissions..."
   chmod -R u+rwX ${GITLAB_DATA_DIR}/tmp/


entrypoint.sh
```diff
@@ -845,8 +845,8 @@ chown ${GITLAB_USER}:${GITLAB_USER} ${GITLAB_BACKUP_DIR}
 
 # create the uploads directory
 mkdir -p ${GITLAB_DATA_DIR}/uploads/
-chmod 0750 ${GITLAB_DATA_DIR}/uploads/
-chown ${GITLAB_USER}:${GITLAB_USER} ${GITLAB_DATA_DIR}/uploads/
+chmod -R u=rwX,g=rX,o= ${GITLAB_DATA_DIR}/uploads/
+chown -R ${GITLAB_USER}:${GITLAB_USER} ${GITLAB_DATA_DIR}/uploads/
```
sameersbn: Here we are only setting permissions on  the topmost `uploads/` directory so that this step does not effect the start up time of the container. If user notices issues, he/she is expected to run `app:sanitize` to fix them

entrypoint.sh
```diff
@@ -978,10 +978,8 @@ appSanitize () {
   chown -R ${GITLAB_USER}:${GITLAB_USER} ${GITLAB_BUILDS_DIR}
 
   echo "Checking uploads directory permissions..."
-  chmod -R 0750 ${GITLAB_DATA_DIR}/uploads/
-  find ${GITLAB_DATA_DIR}/uploads -type f -exec chmod 0644 {} \;
-  find ${GITLAB_DATA_DIR}/uploads -type d -not -path ${GITLAB_DATA_DIR}/uploads -exec chmod 0755 {} \;
-  chown ${GITLAB_USER}:${GITLAB_USER} ${GITLAB_DATA_DIR}/uploads/
+  chmod -R u=rwX,g=rX,o= ${GITLAB_DATA_DIR}/uploads/
+  chown -R ${GITLAB_USER}:${GITLAB_USER} ${GITLAB_DATA_DIR}/uploads/
```
sameersbn: Only here we do a recursive `chown` / `chmod`

The GitLab suggested permission settings, set directories to `0755` and files to `0644`. But your changes sets everything to `750`. Not sure if this could cause any problems.

Allineer:
> so that this step does not effect the start up time of the container.
If we have the `${GITLAB_DATA_DIR}` already (not first container start), by changing permission we can to prevent any notices.

Why not?

sameersbn:
In that case we could run all of the `app:sanitize` commands at every 
start. Issue is when you have a large number of repos or in this case 
uploads, this step will take a long time to complete. This is okay if 
the permissions of an existing installation are busted for some reason. 
But generally this is not going to be the case. So adding this to the 
regular start of the image would been that everyone will have to face 
the increase in startup time.

On Tuesday 13 October 2015 11:56 AM, Allineer wrote:
>     so that this step does not effect the start up time of the container.
>     If we have the |${GITLAB_DATA_DIR}| already (not first container
>     start), by changing permission we can to prevent any notices.
>
> Why not?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sameersbn/docker-gitlab/pull/465#issuecomment-147617842>.
>

-- 
Sameer Naik

w: http://www.damagehead.com e: sameer@damagehead.com

Allineer:
You're right. Thanks.

sameersbn:
Users already complain about the asset compilation step resulting in 
high startup times. Even though this occurs only on the first boot or 
after an upgrade.


On Tuesday 13 October 2015 12:26 PM, Allineer wrote:
> Closed #465 <https://github.com/sameersbn/docker-gitlab/pull/465>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sameersbn/docker-gitlab/pull/465#event-433695278>.
>

-- 
Sameer Naik

w: http://www.damagehead.com e: sameer@damagehead.com

PR #480, opened 2015-10-23
Upgrade to 8.1.0

mikew:
Biggest change was the removal of the whenever gem. The cron stuff in entrypoint.sh failed without it and instead of looking into the issue I opted to install it manually and still run `whenever -w`.


Allineer:
> The cron stuff in entrypoint.sh failed without it

Just drop [this](https://github.com/sameersbn/docker-gitlab/blob/master/entrypoint.sh#L944) line.

sameersbn:
no worries... i have a working copy. I will apply the changes on top of this PR.

PR #508, opened 2015-11-23
Use latest label to replace fixed version

zddhub:
I think use the latest label is better on Upgrade section


sameersbn:
The `latest` build is generated from the master. This is the development build and its use is not always recommended.

When users report issues, it is easier for them to tell the version that they are using which allows us to test with the same version avoiding a lot of back and forth.

I do not see anything good coming out of this change.

zddhub:
I see, thanks

PR #520, opened 2015-11-30
add: script: add signal handler function when use "docker stop" command

dayiguizhen:
When use "docker stop" command to stop container,docker will send SIGTERM to container.And the in this container,the     process whose num is 1 will receive this signal.But if this process didn't do something, docker will kill this container rude. So I add this signal handle function make sure container stop gracefully.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,10 @@ COPY assets/runtime/ ${GITLAB_RUNTIME_DIR}/
 COPY entrypoint.sh /sbin/entrypoint.sh
 RUN chmod 755 /sbin/entrypoint.sh
 
+RUN rm /usr/bin/supervisord
+COPY supervisord /usr/bin/supervisord
+RUN chmod 755 /usr/bin/supervisord
+
 EXPOSE 22/tcp 80/tcp 443/tcp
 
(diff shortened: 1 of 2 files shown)


sameersbn:
Gitlab and related components are not started from the init scripts. They are started by supervisord. 
I need to inspect why gitlab wont shutdown gracefully. It was working fine earlier. Will check and let you know.

dayiguizhen:
I think supervisord monitor  process like gitlab and nginx and so on.

But I'm not sure if supervisord will send SIGTERM signal to gitlab and nginx if supervisord receive SIGTERM(by docker stop).

If supervisord didn't do anything ,docker will send SIGKILL to supervisord .So supervisord  have to shutdown.

Maybe in this time, supervisord didn't have energy to help shutdown gitlab gracefully.

sameersbn:
I have been looking into this issue. But first regarding supervisord. When supervisord receives the `TERM` signal will will send the same signal to all processes that it is supervising. The signal that a process receives can also be configured using the `stopsignal` parameter in the supervisor conf files.

Getting to the issue of graceful termination of the container as it apears, one of the gitlab components `gitlab-workhorse` does not terminate gracefully regardless of the signal being sent to terminate the process. Until this is handled in `gitlab-workhorse` there is nothing we can do to resolve this.

p.s. all the other process terminate gracefully as expected.

Issue #516, opened 2015-11-27
sed error on gitlab startup


JrCs:
When i start my container i have:
<pre>Nov 27 15:30:52 sd-65534 docker[63801]: Configuring gitlab::database
Nov 27 15:30:52 sd-65534 docker[63801]: sed: -e expression #1, char 21: unknown option to `s'
Nov 27 15:30:52 sd-65534 systemd[1]: gitlab.service: Main process exited, code=exited, status=1/FAILURE</pre>
The problem is that you forget to declare the default values of the new variables (GITLAB_SHARED_DIR, GITLAB_ARTIFACTS_DIR, GITLAB_ARTIFACTS_ENABLED) in the Dockerfile

sameersbn:
@JrCs just checked there is a typo that needs to be fixed.

sameersbn:
fixed the typo. can you please check the `latest` tag once the build is ready. Once you confirm I will make a new point release

sameersbn:
nevermind. i made a point release.

JrCs:
Same error using **latest** tag:
<pre>Nov 28 14:09:29 sd-65534 docker[21806]: Configuring gitlab::database
Nov 28 14:09:29 sd-65534 docker[21806]: sed: -e expression #1, char 21: unknown option to `s'
Nov 28 14:09:29 sd-65534 systemd[1]: gitlab.service: Main process exited, code=exited, status=1/FAILURE</pre>

I thing some variables value are missing

sameersbn:
@JrCs It appears that some character in some value like DB_PASS is confusing `sed`. Please add `DEBUG=true` to the environment of the gitlab container and you will see which sed command is failing.

JrCs:
Found, it's DB_PASS that contain a | (your sed separator). Error will be raised if string contain another char use by sed like: $, ( , ), etc...
Perhaps the best is to use:
```bash
envsubst < template.txt
```
which work even if variables contain specific chars.
But you need to rewrite your templates from {{VAR}} to ${VAR}

sameersbn:
this change looks massive in the way it would alter the setup procedure.
why not just change the password?

JrCs:
I can but if someone define a pass containing things like $ ( ) [ ] | etc... the error will appear.
Perhaps it's better to find a definitive solution ?

sameersbn:
[...]

PR #522, opened 2015-12-01
Fix problems to replace some placeholder values with sed

### LINKED ISSUE #408: Install fails when postgres password has / in it
The container keeps exiting because some sed statements are failing.  The sed statements fail with:
sed: -e expression #1, char 19: unknown option to `s'

I discovered this was the root cause by adding DEBUG_ENTRYPOINT into my environment.

P.S.  No rush on this issue.  I am changing the password to my postgres user.

JrCs:
Keep the same template files but use envsubst instead of sed to replace placeholder values.

Resolves #408 and #516

Use a generic function (replace_placeholders) to make the job.


sameersbn:
I see you made the changes for all configurations. I test locally and let you know. thanks for the PR.

sameersbn:
Have merged upstream changes and updated to use replace_placeholders (renamed to update_template) whenever possible. You can take a look at the changes here https://github.com/JrCs/docker-gitlab/compare/master...sameersbn:JrCs-pr-updates. I am currently testing these changes on my local install and will merge the PR after I am confident that it will not lead to breakages :smile:

sameersbn:
was also thinking of updating `update_template` to automatically select the user for executing `envsubst` based on the ownership of the target file. This will allow us to use `update_template` also for the nginx configuration updates

JrCs:
Good idea

PR #545, opened 2015-12-21
Implement unix socket connection to redis

### LINKED ISSUE #415: Support UNIX sockets for db connections
Reading [the documentation](https://hub.docker.com/r/sameersbn/gitlab/) I cannot find how to configure GitLab docker container to connect to a database (in my case: PostgreSQL) via a UNIX socket. If UNIX sockets are not supported, it would be worthwhile to add such a support, methinks!

ivlis:
This PR is aimed to close #415 

It is the first step and I want some feedback, if I am going in right direction.


sameersbn:
Couple of notes:
 - adding `-v /run/redis` to `sameersbn/redis` will expose the redis socket when `--volumes-from redis` is used.
 - we don't want these settings by default in the `docker-compose.yml`, IMO this is an advanced user thing. Support wise this could be more than what we ask for.
 - instead of `REDIS_CONNECTION=unix`, internally we can check for the existence of `/run/redis/redis.sock` and configure accordingly. This can be done in the `gitlab_finalize_redis_parameters` where we perform the discovery from docker links. Here we can check the existence of the socket, or fallback to tcp provided that `REDIS_HOST`/`REDIS_PORT` are not specified by the user. Precedence should be:
   - user config
   - unix socket
   - discovery from docker the `redisio` link alias

Anyways, this is a small enough change. If you are OK with me taking over this PR from this point, I will be happy to do it over the weekend.

ivlis:
@sameersbn Thanks, I'll do that, as well as for PostgreSQL, mySQL and nginx. Now I know the pattern.

Also, I'm going to separate the nginx server to a separate container (probably in another PR), what do you think?

sameersbn:
can you please create separate PR's for each.

ivlis:
OK. And I'll update for 8.3.0 also.

ivlis:
DONE

PR #548, opened 2015-12-23
Implement unix socket connection to redis

ivlis:
This is the first step to create a full unix socket based scheme of
communications between the related containers.

Sockets-dvc is a new DVC that will contain all sockets.

The socket connection is used if the linked redis container is used and
a socket is available inside the docker container. Otherwise, we fall
back to a tcp connection.

Fixes #548


ivlis:
@sameersbn any plans for merging this?

PR #550, opened 2015-12-23
mysql over socket

ivlis:


assets/runtime/functions
```diff
@@ -75,8 +75,16 @@ gitlab_finalize_database_parameters() {
   # port 3306 and 5432 respectively.
   if [[ -n ${MYSQL_PORT_3306_TCP_ADDR} ]]; then
     DB_ADAPTER=${DB_ADAPTER:-mysql2}
-    DB_HOST=${DB_HOST:-${MYSQL_PORT_3306_TCP_ADDR}}
-    DB_PORT=${DB_PORT:-${MYSQL_PORT_3306_TCP_PORT}}
+    sleep 5  # wait till MySQL creates a socket
```
robinro: `sleep 5`?
This should not run by default.
Make a variable to ask whether we look for the socket or not and if we do, do some quicker look to check for it which aborts after 5 secs.

Issue #521, opened 2015-12-01
issues with missing icons (assets) via using GITLAB_RELATIVE_URL_ROOT


BeneDicere:
Hi,
version 8.2 worked well but since version 8.2.1 icons in side-bar and header are not showing up ![sidebar](https://cloud.githubusercontent.com/assets/5327850/11495698/93812926-980d-11e5-8727-b11ac3bb7ad5.png).
The JavaScript console shows that it is ignoring the relative url and looking for something in ROOT/assets/... I don't know whether this is a GitLab, Docker Image or Apache issue, but because it worked well until now i hope its a Docker Image issue. :)

Thanks for the Image, until now it worked very well and made the deployment very easy.

sameersbn:
I do not see this issue. For the test I used the sample `docker-compose.yml` file with `GITLAB_RELATIVE_URL_ROOT=/gitlab`

sameersbn:
Are you using apache as a load balancer?

BeneDicere:
we are using apache as reverse proxy, thats all we have configured:

  ProxyRequests Off
  ProxyPass /gitlab http://localhost:10080/gitlab
  <Location /gitlab>
    ProxyPassReverse http://localhost:10080/gitlab
  </Location>

Strange, because beside of the gitlab image (it was updated) nothing changed. I will see whether i can find another reason for this behaviour.
Thanks for your fast answer.

sameersbn:
can you try connecting directly to `http://localhost:10080/gitlab` to verify in the urls are correct?
p.s. i think `ProxyPassReverse http://localhost:10080/gitlab` should be `ProxyPassReverse http://localhost:10080`. not a 100% sure.

BeneDicere:
bypassing the proxy works. i will see what else the reason for this issue is, strange that this did not happen before. thanks for the pointer, sorry for the noise.

sameersbn:
In case you have not tried already, please clear the browser cache and retry.

BeneDicere:
This was done with different browsers and of course a cache clear. ;) Beside of bypassing the proxy the current workaround is to add a separate proxy rule for the "/asset" path to "/gitlab/asset". But i think this is  a local and no general issue.
Once again, thanks for this great image and your fast help.

mlewe:
I have the same problem. In the generated CSS files the wrong asset path is referenced. E.g.:
[...]

PR #555, opened 2015-12-24
Segregate nginx to a separate container

ivlis:
Following isolation principles of docker the nginx process is moved to a separate
container, instead running in the gitlab container. The nginx container uses an
official nginx image. The configuration scripts detects if an nginx container is
linked and does not start nginx inside the gitlab container in that case. If no
nginx container is detected the script works as normal.

Important changes:
* The gitlab-workhorse socket is moved to a more reasonable location
* Variable `NGINX_USE_EXTERNAL`. If true, the gitlab container does not start
  an nginx process, even if an nginx container is not detected. Useful if one have
  an nginx process outside of docker-compose


veggiemonk:
Nice! 

> The configuration scripts detects if an nginx container is
linked and does not start nginx inside the gitlab container in that case. If no
nginx container is detected the script works as normal.

This is so convenient! Thanks a lot.

ivlis:
#521 blocks this. It is better to wait till it will be fixed in the upstream gitlab-workhorse project. It is not possible to configure unicorn (or I don't know how) to open a socket at a location other from `tmp/sockets` and I don't want to mess with symlinks.

BTW, why not allow nginx to serve static files directly, comes as a great mystery to me.

sameersbn:
@ivlis when using relative_urls gitlab rebuilds some of the assets to correct the urls. serving users the static assets using nginx would result in broken links for users who choose to use relative urls. when not in use, it is perfectly fine for nginx to serve these prebuilt assets.

Also I would like to mention that recently there has been a lot of activity in GitLab with regards to handing of http requests, wherein they are working towards offloading all http request handling to `gitlab-workhorse`. I believe that when this transition is complete nginx will become completely optional at which point we might would be able to drop it from the image altogether.

Given this development activity, I think we should avoid merging this change in the image, because with every release things will start breaking on the http request handling front and during this time it is best to stick with how gitlab inc. recommends installing Gitlab and finally make the transition when things are ready.  See "Decoupling GitLab from NGINX" section of [GitLab 8.3.0 Release Announcement](https://about.gitlab.com/2015/12/22/gitlab-8-3-released/)

MeinAccount:
What is the current state about separating Nginx? [All requests are handled by `gitlab-workhorse` these days.](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/lib/support/nginx/gitlab)

ngpestelos:
Can we close this PR? @ivlis @sameersbn

SolidNerd:
In my opinion we can close this, because we got #794 . And you don't need to communictae with nginx you can communicate with gitlab workhorse over `8181` that is possible since #793 .

SolidNerd:
I will close this now because we got #794 and a schedule for it.

PR #573, opened 2016-01-14
FIx CVE-2016-0777 and CVE-2016-0778

lorenz:
Applies the proposed workaround from http://undeadly.org/cgi?action=article&sid=20160114142733 as Gitlab uses an SSH client to fetch resources from remote Git servers

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,8 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv E1DD270288B4E60
  && gem install --no-document bundler \
  && rm -rf /var/lib/apt/lists/*
 
+RUN echo -e 'Host *\nUseRoaming no' >> /etc/ssh/ssh_config # Fix CVE-2016-0777 and CVE-2016-0778
+
 COPY assets/build/ ${GITLAB_BUILD_DIR}/
 RUN bash ${GITLAB_BUILD_DIR}/install.sh
 


Dockerfile
```diff
@@ -38,6 +38,8 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv E1DD270288B4E60
  && gem install --no-document bundler \
  && rm -rf /var/lib/apt/lists/*
 
+RUN echo -e 'Host *\nUseRoaming no' >> /etc/ssh/ssh_config # Fix CVE-2016-0777 and CVE-2016-0778
+
```
sameersbn: please move this configuration to after https://github.com/sameersbn/docker-gitlab/blob/master/assets/build/install.sh#L131

sameersbn:
btw, it appears that a new openssh point release was released by debian/ubuntu packaging, which probably plugs the security vulnerability. This is available in the `8.3.4-1` hotfix. This PR may not be required.

lazyfrosch:
The Ubuntu update should fix the problem. Also I'm wondering why a problem in the ssh *client* would cause us trouble here?

sameersbn:
@lazyfrosch i dont think the issue is specific to the client. maybe I am wrong.

lazyfrosch:
@sameersbn it is a bug in the SSH client with a half implemented feature, that can be broken by a malicious server, where the client might leak private keys in its memory.

lorenz:
According to [USN2869](http://www.ubuntu.com/usn/usn-2869-1/) it is now fixed in Ubuntu, so this PR is no longer necessary. Only the base image needs to be updated.

PR #600, opened 2016-02-10
Gitlab Env's in a  seperate file

solidnerd:
Hey, 
to make the environment configuration simpler. You could use an .env file. In my opionion it's a better way to set the environment configuration of the gitlab container. You could use it with the `docker run --env-file=./.env` aswell.  Let me know what you think about it.


sameersbn:
I don't think this is a good idea. It will break the quick start guide and will be painful to explain to users. Right now the quick start guide is plain and simple, download the compose file, edit and run.

sameersbn:
Instead of the proposed solution I suggest loading an `.env` file from the data volume if it exists. ie. at https://github.com/sameersbn/docker-gitlab/blob/master/assets/runtime/env-defaults#L4 we add

```bash
if [[ -f ${GITLAB_DATA_DIR}/.env  ]]; then
  source ${GITLAB_DATA_DIR}/.env
fi
```

User would be able to create a `.env` file in the data volume with the desired configuration and would override anything that's specified using `--env` arguments.

I really don't want to get rid of the configuration options from docker-compose.yml since it's the expected location to specify env variables. the `.env` option could be a hidden or advanced configuration option for advances/experienced users :smile: 

What do ya think?

Estebanv:
I had some issues that could be fixed with the env file. Check this out: [https://github.com/docker/compose/issues/2120](url)
My directory structure is:
/docker/gitlab/src
/docker/gitlab/data
/docker/redmine/src
/docker/redmine/data

For the yml and volumes. I'm also using your redmine containers, so this creates a conflict. With the env file, you can use a default COMPOSE_PROJECT_NAME to prevent that. I know you can also use the '-p' parameter, but the you have to use it with every command (up, stop, rm, etc).
Also i think this will lead to a much cleaner yml, which make its easier to read. I dont see how this breaks the guide, its the same in another file.

> I really don't want to get rid of the configuration options from docker-compose.yml since it's the expected location to specify env variables.
Actually, the env file is where they are supposed to be IMHO.
Thanks for the hard work btw.

solidnerd:
I will close this. This isn't needed anymore.

PR #629, opened 2016-03-06
Use Ruby 2.2

splattael:
Is there any reason not to use Ruby 2.2 with GitLab?

I'm testing it right now and am not seeing any errors.

WDYT?


sameersbn:
The GitLab Install Docs states that currently only ruby `2.1.x` is supported, see https://github.com/gitlabhq/gitlabhq/blob/master/doc/install/installation.md#2-ruby

Once GitLab announces support for `2.2.x` or above we can update to the supported version.

splattael:
Yep, you're right. They still have some issues with Ruby 2.2:
https://gitlab.com/gitlab-com/www-gitlab-com/merge_requests/1434#note_3826391

I'll keep testing 2.2 though :)

Thanks!

PR #638, opened 2016-03-16
Update GitLab to 8.5.7

stuwil:
Bumps minimum git version to avoid remote code execution vulnerability, see https://about.gitlab.com/2016/03/16/gitlab-8-dot-5-dot-7-released/


stuwil:
Available as stuwil/docker-gitlab:8.5.7 from Docker Hub if you can't wait.

PR #655, opened 2016-04-06
fix invalid field TCPSocket for v1.Probe.

xueshanf:


xueshanf:
Would you like to update these pods yml file with ReplicationController kind? I can do a RP if you think it's a good idea.

sameersbn:
Yes sure.

PR #658, opened 2016-04-07
Docker compose v2

18601673727:
**Purpose:**
Solving https://github.com/sameersbn/docker-gitlab/pull/631 and https://github.com/sameersbn/docker-gitlab/issues/630.

**Known issue:**
- <del>SSH server seems not working, only be able to pull/push via HTTPS. </del>
- <del>HTTP(S) clone url which shown in the "project detail page" is not correct. (only with https://github.com/sameersbn/docker-gitlab/pull/658/commits/222267976b888ddd403161561f5f645761177ba6)</del>

PR #661, opened 2016-04-09
finalize `docker-compose.yml` for v2

sameersbn:


sameersbn:
Ideally I would have liked to use a `volumes` section to define named volumes. but unfortunately the `local` volume driver does not allow specifying a host path for the volume.

I have also applied the `:Z` [volume label](https://docs.docker.com/engine/userguide/containers/dockervolumes/#volume-labels), but haven't tested it on a selinux based distro like redhat or centos. It would be great if someone could test this.

SolidNerd:
This would possible if you use a volume plugin like [local-persist](https://github.com/CWSpear/local-persist)

sameersbn:
Thanks for linking to `local-persist`. Good to know :smile: 
For the sample docker-compose.yml file we need to stick with what docker provides as the default.

SolidNerd:
@sameersbn. I put my test results in here as you mentioned me in #658 . I guess it will provide a litte bit more clarity. 
First of all everything is working for me now. Thanks for the work. But i got some  barriers. 

1. If you don't have the latest redis image you get an error message that the command does not work.
   To fix it. You to pull the latest redis image.
2. Maintenance has been changed. The way of restoring a backup is different to the old way, because your gitlab container must be in the same network and you should provide the enviroment variables to ensure that the gitlab_container finds the database.
To fix it i use the following command: 
```
docker run --name gitlab -it --rm --net="gitlab_gitlab_backend" -v /srv/gitlab/gitlab:/home/git/data --env-file=gitlab.env sameersbn/gitlab:8.6.4 app:rake gitlab:backup:restore BACKUP=1460347212
```
I used the env variables from the compose file. In that case we sould use env_file to define them once. Lets rethink about #600.

sameersbn:
> If you don't have the latest redis image you get an error message that the command does not work. To fix it. You to pull the latest redis image.

I guess the redis client version in the gitlab image may not be compatible. 

> Maintenance has been changed. The way of restoring a backup is different to the old way, because your gitlab container must be in the same network and you should provide the enviroment variables to ensure that the gitlab_container finds the database. To fix it i use the following command:

Wouldn't you use docker-compose to launch the rake task?

# create a backup

```bash
docker-compose run --rm gitlab app:rake gitlab:backup:create
```

# restore a backup

```bash
docker-compose run --rm gitlab app:rake gitlab:backup:restore BACKUP=1460905799
```

> Lets rethink about #600.

I have commented on the issue https://github.com/sameersbn/docker-gitlab/pull/600#issuecomment-211043277

SolidNerd:
I mean `sameersbn/redis:latest` image. I needed to pull that than everything works at `docker-compose up`. 

Next time i will try to use docker-compose to run the rake tasks it will make my work a lot of easier :sweat_smile: .

sameersbn:
> I mean sameersbn/redis:latest image. I needed to pull that than everything works at docker-compose up.

Yes. I meant the `redis-cli` version in the gitlab image was probably not compatible with the `redis-server` from `sameersbn/redis:latest`

> Next time i will try to use docker-compose to run the rake tasks. It will make my work a lot of easier. :sweat_smile:

With docker-compose v2 I think this might be the only valid way of doing things correctly because of all the networking bits.

docker-compose.yml
```diff
@@ -1,34 +1,59 @@
-postgresql:
-  restart: always
-  image: sameersbn/postgresql:9.4-18
-  environment:
+version: '2'
+
+networks:
```
SolidNerd: The networks key isn't needed because docker-compose will create an own network by default.

jgsqware:
Hello, any idea when you will merge it?

docker-compose.yml
sameersbn: Yep, its not "required". 
Not sure if we should keep it or remove it. Guess we can remove it for now and add it later as we see fit?

sameersbn:
@jgsqware can you try it out and let us know it works for you? will merge it on the weekend after doing some tests to check if backup/restore operations work without issues.

PR #679, opened 2016-04-24
Use official docker in docker image

splattael:
Since 1.10 CI runner does not require dind to have a bash installed anymore.

From now on, we can use the official dind image.


splattael:
Thanks for merging and for your tireless effort maintaining this awesome Docker image :green_heart: :heart: :blue_heart: :smiley:

PR #684, opened 2016-05-05
Configure SAML attribute statements and external groups

patope:
Configuration parameters for saml attribute statements and external groups.


sameersbn:
@patope thanks for the PR. I am unable to test any SAML stuff other then verify that the parameters are being set. Can you make sure this is tested well on your end?

patope:
Example settings `groups_attribute: 'Groups'` and `external_groups: ['Contractors', 'Freelancers']` broke saml login for me, so wanted to get rid of those.
I use this with custom attribute statements in production. I also tested default case too, where attribute_statement is not added (uses to omniauth's default attributes). Group settings generates sensible config and gitlab starts. But, my saml idp do not send groups, so i could not really validate it.

I'm quite confident this is ok...

sameersbn:
Need to figure out a way to add the groups in quotes as in `[ 'Contractors', 'Freelancers' ]`. I think that its important to make sure the group names can have spaces. Not sure it its permitted in the SAML config though

patope:
docker-compose.yml env settings
```
    - OAUTH_SAML_GROUPS_ATTRIBUTE=external-groups
    - OAUTH_SAML_EXTERNAL_GROUPS='group1','group2'
```
generates
```
      - { name: 'saml',
          label: 'Our SAML Provider',
          groups_attribute: 'external-groups',
          external_groups: ['group1','group2'],
          args: {
```
for me...

sameersbn:
in that case, i think we should quote an example in the readme. what do you think?

xueshanf:
+1. Thanks for the fix! 8.7.2 broke my SAML login b/c our IdP  has not setup to deal with the new external groups settings. Yay!

Issue #595, opened 2016-02-05
where can i find gitlab-ctl ? 


lgersman:
Hello,

i actually want to import repositories from another gitlab instance into my docker-gitlab instance.

As far as i read i have to do a bunch of "gitlab-ctl" calls (http://stackoverflow.com/questions/19902417/change-the-data-directory-gitlab-to-store-repos-elsewhere) but I cannot find gitlab-ctl in the docker container.

Do I have to install it additionally ? 

Kind regards,

Lars

sameersbn:
This image is built from GitLab sources. `gitlab-ctl` is part of the the 
omnibus packages and as such is not available in here.

On Friday 05 February 2016 06:33 PM, lgersman wrote:
> Hello,
>
> i actually want to import repositories from another gitlab instance into
> my docker-gitlab instance
>
> As far as i read i have to do a bunch of "gitlab-ctl" calls
> (http://stackoverflowcom/questions/19902417/change-the-data-directory-gitlab-to-store-repos-elsewhere)
> but I cannot find gitlab-ctl in the docker container
>
> Do I have to install it additionally ?
>
> Kind regards,
>
> Lars
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sameersbn/docker-gitlab/issues/595>.
>

-- 
Sameer Naik

w: http://www.damagehead.com e: sameer@damagehead.com

veggiemonk:
Hi @lgersman 

Had the same problem a while back. There are actually many ways to import repositories.

1. Import by the web interface (one by one) from many 3rd parties hosting repo. 
2. Import by file system (batch import of many repositories) <= I guess this is what you are looking for ?
3. Sync repo (mirroring is available in EE version only but you can manage it with scripts, although it would be wise to really know `git` internals before going on such a journey). 

Assuming that you want to batch-import many repos. 
The fatest and simplest way is to copy the bare repos to the gitlab volume. I'll explained, the default config of the `docker-compose.yml` places GitLab repos in `/srv/docker/gitlab/gitlab/repositories` on the host, they are in `/home/git/data` inside the container. 
Copy all those bare repos  (by `rsync` or `scp`, `rsync` is faster) to that folder on the host.
Before you run the next command, make sure the DB container (`postgres` by default) and the `redis` container are running
Then you need to import them into GitLab by running a `rake` command

```
docker run --name gitlab -it --rm \
    --link redisio:NAME_OF_REDIS_CONTAINER \
[...]

PR #690, opened 2016-05-13
Explain [OPTIONS] placeholder in docker run commands

debovema:
Based on discussion in #595, a comprehensive explanation on how to use "docker run" commands to execute Rake tasks for instance.


debovema:
up

PR #692, opened 2016-05-17
Upgrade to 8.8.0-rc1

SolidNerd:
Hey, 
i would like to know if you want to support rc's. I tested it on my GitLab and it works for me. I want to know what are you thinking about it.


sameersbn:
I think if we want to support RC's then we should probably create a new development branch. adding RC releases to the master could be risky if a user unknowingly pulls the `latest` tag from the dockerhub. We could create a development tag as well.

SolidNerd:
I agree with you. So can you create the branch and  i will make a new PR.

sameersbn:
I am just wondering how we would want the development to continue. should we do all development on the development branch and merge with master on new releases?

PR #697, opened 2016-05-22
Upgrade to 8.8.0

SolidNerd:
I made a PR for it, but i guess it's not released yet only the tag was made. So we need to wait with this.

**EDIT:**
For everyone who wants to try it out. I have been created a [Gist](https://gist.github.com/SolidNerd/1bf47f85457e4f90eea541586f7290da) for all configurations that are needed.


SolidNerd:
In my opinon we could merge it. An Official [Announcement](https://about.gitlab.com/2016/05/22/gitlab-8-8-released/) is published.

Ping @sameersbn

sameersbn:
@SolidNerd there are a few additional changes that need to be done. I am working on adding those. Will merge with the changes soon

SolidNerd:
Okay. It seems an admin bug exists currently. So we should wait until this [MR](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/4242) is merged.

sameersbn:
@SolidNerd oh.. thanks for pointing it out. I was trying to debug that.

SolidNerd:
So any things left to do for the update ?

sameersbn:
That seems to be some other issue.
I dont see the admin user (`root`) being created on first boot.

sameersbn:
This is the one major issue I see making the release.
Other things are pretty basic like updating the `gitlab.yml` config and nginx configuration templates.

SolidNerd:
The admin bug will be fixed in 8.8.2 it's  mentioned in the [Changelog](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/CHANGELOG)

SolidNerd:
@sameersbn  I added all related configuration options for the container registry. Tomorrow i will look at the nginx configuration template. After this is finished I will try to create the admin user. If you found more related stuff for the release.

aranw:
How will this work with GitLab Registry Container? Will there be examples on how to link them together?

README.md
```diff
@@ -821,7 +821,7 @@ Below is the complete list of available options that can be used to customize yo
 - **GITLAB_SSH_PORT**: The ssh port number. Defaults to `22`.
 - **GITLAB_RELATIVE_URL_ROOT**: The relative url of the GitLab server, e.g. `/git`. No default.
 - **GITLAB_TRUSTED_PROXIES**: Add IP address reverse proxy to trusted proxy list, otherwise users will appear signed in from that address. Currently only a single entry is permitted. No defaults.
-- **GITLAB_REGISTRY_ENABLED**: Enables the GitLab Container Registry. Defautls to `true`.
+- **GITLAB_REGISTRY_ENABLED**: Enables the GitLab Container Registry. Defautls to `false`.
```
lenovouser: Typo here, called `Defaults`

README.md
```diff
@@ -796,6 +796,7 @@ Below is the complete list of available options that can be used to customize yo
 - **GITLAB_PROJECTS_WIKI**: Set if *wiki* feature should be enabled by default for new projects. Defaults is `true`.
 - **GITLAB_PROJECTS_SNIPPETS**: Set if *snippets* feature should be enabled by default for new projects. Defaults is `false`.
 - **GITLAB_PROJECTS_BUILDS**: Set if *builds* feature should be enabled by default for new projects. Defaults is `true`.
+- **GITLAB_PROJECTS_CONTAINER_REGISTRY**: Set if *container_registry* feature should be enabled by default for new projects. Defaults is `true`.
```
lenovouser: Typo

```diff
+Default
-Defaults
```

or

```diff
+to
-is
```

README.md
```diff
@@ -820,13 +821,22 @@ Below is the complete list of available options that can be used to customize yo
 - **GITLAB_SSH_PORT**: The ssh port number. Defaults to `22`.
 - **GITLAB_RELATIVE_URL_ROOT**: The relative url of the GitLab server, e.g. `/git`. No default.
 - **GITLAB_TRUSTED_PROXIES**: Add IP address reverse proxy to trusted proxy list, otherwise users will appear signed in from that address. Currently only a single entry is permitted. No defaults.
+- **GITLAB_REGISTRY_ENABLED**: Enables the GitLab Container Registry. Defautls to `false`.
```
lenovouser: Typo

```diff
+Defaults
-Defautls
```

SolidNerd:
@aranw Thanks for the question. I didn't test it completely. After the test is finished I will try to write a guide for that. I guess the best place to start will be the [GitLab Container Registry Administartion Documentation](http://docs.gitlab.com/ce/administration/container_registry.html) . 

The communication should be implemented with `docker networks`, because we have only a `docker-compose.yml ` `V2` and this uses `docker networks`. The second argument would that docker marks the `--link` API as deprecated since Docker 1.9 and I don`t know when they drop it completely . In my opion the only way would be with `docker networks`.

README.md
```diff
@@ -820,13 +821,22 @@ Below is the complete list of available options that can be used to customize yo
...
+- **GITLAB_REGISTRY_HOST**: Sets the Gitlab Registry Host. Defaults to `registry.example.com`
+- **GITLAB_REGISTRY_PORT**: Sets the GitLab Registry Port. Defaults to `5000`.
+- **GITLAB_REGISTRY_API_URL**: Sets the Gitlab Registry Api URL. Defaults to `http://localhost:5000`
```
lenovouser: Typo

```diff
+API
-Api
```

vitorarins:
Hi, will this be merged after release `8.8.2` from Gitlab? Or will it be merged now and another PR will update to `8.8.2`? I am sorry for my confusion, I don't know how are you guys working with versions here. Thank you for the great work!

sameersbn:
@vitorarins I think we'll directly release `8.8.2` or whichever version fixes the admin account creation.

sameersbn:
@SolidNerd do you see the admin account creation issue?

SolidNerd:
@sameersbn Yes. It's needed to be fixed. I guess we missunderstood us, because i mean with the admin bug another problem. My admin bug was when you enter the `admin/application_settings`  you get an 500.

sameersbn:
@SolidNerd the admin account creation issue seems to be something that is specific to this image. Tried `gitlab/gitlab-ce:8.8.1-ce.0` and had no problems. Will need some time to debug it.

SolidNerd:
@sameersbn Okay. In the meatime I will try to test the container registry feature completly to ensure that it works or problems are by gitlab itself.

mgansler:
How will the registry work if gitlab itself is running behind a reverse proxy?

SolidNerd:
@mgansler At the moment I don't know. My production setup has a reverse proxy. After i finished all tests i will try to make a duplicate of my production system and test it with this setup. But this needs some time. I guess that I have finished the tests tomorrow in the afternoon and I will report you about my results.

lenovouser:
@sameersbn `v8.8.2` has been released by the way. So this should be good to merge, right?

boonkerz:
i have no registry entry in my project settings
-e 'GITLAB_REGISTRY_ENABLED=true'
should be enough?

SolidNerd:
@lenovouser Currently an issue exist that you can't create an admin account with a fresh version of 8.8.0 this must be fixed before we can merge it.

@bboehmke You need to configure all related `GITLAB_REGISTRY_*` env's.

boonkerz:
@SolidNerd ok because but why the defaults?
it looks like they are preset :)

SolidNerd:
@boonkerz You are right. It's currently a bug that the variables aren't subsituted correctly. I will fix this now..

hanej:
Thank you @sameersbn and @SolidNerd for your hard work on this image.  It's greatly appreciated.

SolidNerd:
Some updates on my progress for the container registry. Currently I have problems to run the docker-registry. It throws me  some cruel ssl errors but i use an offical ssl certificate.

robertoamoreno:
Also would be handy to have S3  Backend ENV for Registry

athieriot:
Gitlab doesn't seem to support S3 for Docker Registry at the moment :( It needs the filesystem path to the image storage

vitorarins:
@SolidNerd the errors you are getting are on deploy? Or when you are using the registry? If it is on deploy, you could have the wrong order of chain in your certificates.

lenovouser:
@SolidNerd any idea on how to fix the admin account `500` bug? I really want to use the Container Registry :3

SolidNerd:
@lenovouser Sorry I have no idea. But I have a workaround for this. Use the `8.7.6` and create with this     an admin account. After that you can use the volumes for 8.8+ .

sameersbn:
@SolidNerd I have figured the admin account creation issue. Can you update the PR with the following patch

```diff
From 8ec557711818e3aa6ee4109d39465d49f036ed03 Mon Sep 17 00:00:00 2001
From: Sameer Naik <sameer@damagehead.com>
Date: Fri, 27 May 2016 22:05:51 +0530
Subject: [PATCH] do `gitlab:setup` after `db:migrate`

---
 assets/runtime/functions | 13 +++++++------
 1 file changed, 7 insertions(+), 6 deletions(-)

diff --git a/assets/runtime/functions b/assets/runtime/functions
index b1d10d1..1b97220 100644
--- a/assets/runtime/functions
+++ b/assets/runtime/functions
@@ -1242,12 +1242,6 @@ migrate_database() {
       COUNT=$(PGPASSWORD="${DB_PASS}" psql -h ${DB_HOST} -p ${DB_PORT} -U ${DB_USER} -d ${DB_NAME} -Atw -c "${QUERY}")
       ;;
   esac
-  if [[ -z ${COUNT} || ${COUNT} -eq 0 ]]; then
-    echo "Setting up GitLab for firstrun. Please be patient, this could take a while..."
-    exec_as_git force=yes bundle exec rake gitlab:setup \
-      ${GITLAB_ROOT_PASSWORD:+GITLAB_ROOT_PASSWORD=$GITLAB_ROOT_PASSWORD} \
-      ${GITLAB_ROOT_EMAIL:+GITLAB_ROOT_EMAIL=$GITLAB_ROOT_EMAIL} >/dev/null
-  fi

   # migrate database if the gitlab version has changed.
   CACHE_VERSION=
@@ -1288,6 +1282,13 @@ migrate_database() {
     rm -rf ${GITLAB_TEMP_DIR}/GITLAB_RELATIVE_URL_ROOT # force cache cleanup
   fi

+  if [[ -z ${COUNT} || ${COUNT} -eq 0 ]]; then
+    echo "Setting up GitLab for firstrun. Please be patient, this could take a while..."
+    exec_as_git force=yes bundle exec rake gitlab:setup \
+      ${GITLAB_ROOT_PASSWORD:+GITLAB_ROOT_PASSWORD=$GITLAB_ROOT_PASSWORD} \
+      ${GITLAB_ROOT_EMAIL:+GITLAB_ROOT_EMAIL=$GITLAB_ROOT_EMAIL} >/dev/null
+  fi
+
   # clear cache if relative_url has changed.
   [[ -f ${GITLAB_TEMP_DIR}/GITLAB_RELATIVE_URL_ROOT ]] && CACHE_GITLAB_RELATIVE_URL_ROOT=$(cat ${GITLAB_TEMP_DIR}/GITLAB_RELATIVE_URL_ROOT)
   if [[ ! -f ${GITLAB_TEMP_DIR}/GITLAB_RELATIVE_URL_ROOT || ${GITLAB_RELATIVE_URL_ROOT} != ${CACHE_GITLAB_RELATIVE_URL_ROOT} ]]; then
--
2.8.3
```

Apple patch with `git am <patch-name>`

Will start testing the PR tomorrow before landing..

vitorarins:
@SolidNerd: I am running a private Registry with [Portus](https://github.com/SUSE/Portus), and thought that I would get rid of all that configuration steps with Gitlab Registry. I just hope it gets a little easier :disappointed:

Ilya-Kuchaev:
Any success here?

sameersbn:
The changes are being tested. Should land soon. Please be patient.

On 29/05/16 5:20 PM, Kuchaev Ilya wrote:
> Any success here?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/sameersbn/docker-gitlab/pull/697#issuecomment-222356569>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAZCI1ihgxpMZZG6AFRyxe_OgXzi76_Bks5qGX17gaJpZM4Ij-Mc>.
>

-- 
Sameer Naik

w: http://www.damagehead.com e: sameer@damagehead.com

SolidNerd:
At all. I can say that the container registry is working now. I guess that I have found  a bug in GitLab this causes the issues but i have fixed it locally. My next plan was to test it all and after that i will cleanup my commits and rebase it.

SolidNerd:
So I'm done. Let's wait for @sameersbn .

SolidNerd:
I have create all relevant PR's #705 and  #704. So lets wait

mgansler:
Great work! I'm going to try it now, just can't wait any longer ;)
What is your setup with the reverse proxy? Does it terminate SSL for your Gitlab instance? What about the registry?

SolidNerd:
Currently I haven't any  computer in my near. I clould you post all related
configurations for the container registry later.

mgansler:
That would be great. Right now I have some issues connection, the registry keeps telling me `400: The plain HTTP request was sent to HTTPS port`

${GITLAB_LOG}/nginx/gitlab_registry_error.log tells me:

```
2016/05/29 18:44:09 [error] 543#543: *1182 connect() failed (111: Connection refused) while connecting to upstream, client: 127.0.0.1, server: registry.domain.tld, request: "GET /v1/users/ HTTP/1.1", upstream: "http://[::1]:5000/v1/users/", host: "registry.domain.tld:5000"
```

Output of `netstat -tulpen` inside the container:
```
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       User       Inode       PID/Program name
tcp        0      0 127.0.0.11:47717        0.0.0.0:*               LISTEN      0          10542054    -               
tcp        0      0 0.0.0.0:5000            0.0.0.0:*               LISTEN      0          10544484    538/nginx -g daemon
tcp        0      0 127.0.0.1:8080          0.0.0.0:*               LISTEN      1000       10544631    -               
tcp        0      0 0.0.0.0:80              0.0.0.0:*               LISTEN      0          10544482    538/nginx -g daemon
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      0          10542874    539/sshd        
tcp6       0      0 :::80                   :::*                    LISTEN      0          10544483    538/nginx -g daemon
tcp6       0      0 :::22                   :::*                    LISTEN      0          10542875    539/sshd        
udp        0      0 127.0.0.11:33746        0.0.0.0:*                           0          10542053    -               
```

If I understand this correctly, it tries to connect via IPv6 but there is nothing listening on `:::5000`.
So I took a look at the nginx configurations for gitlab and for the registry and indeed, there is a difference.
gitlab looks like this:
```
listen 0.0.0.0:80 default_server;
listen [::]:80 default_server;
```

while gitlab-registry looks like this:
```
listen *:{{GITLAB_REGISTRY_PORT}} ssl http2
```

Maybe I'm totally wrong here but I think that the registry is listening only on IPv4 while it tries to connect via IPv6. Will investigate further....

mgansler:
I managed to get rid of the `Connections refused` errors by setting the `GITLAB_REGISTRY_API_URL` to http://gitlab:5000 (where gitlab is the name of my service in the `docker-compose.yml` file).
But the `400 Bad Request: The plain HTTP request was sent to HTTPS port` is still present.
I'm giving up for today, maybe one of you guys has an idea :)

mgansler:
Soon...I didn't gave up and fixed the issue. Too embarrassing to say what it actually was :sob:

nachokb:
@mgansler please... what was it?

blandes:
I'm going to guess http://gitlab:5000 needed to be https://gitlab:5000? @mgansler @nachokb

mgansler:
http was fine but it has to point to the registry, not gitlab itself.

SolidNerd:
Sorry I didn't have yesterday time to do it. I created a [Gist](https://gist.github.com/SolidNerd/d98e940444cbcfbb23c677d285dde8c2) for it. I hope it helps everyone that want's to run the container registry. Post there if you have related issues or questions.

mgansler:
Thanks for that! I actually have a question: What is the GITLAB_REGISTRY_KEY_PATH used for?

SolidNerd:
@mgansler  `GITLAB_REGISTRY_KEY_PATH` is the private key of the `rootcertbundle`. So that GitLab can decrypt all messages from the the docker registry.

mgansler:
I'm thinking about running the registry behind a reverse proxy.
Does this pub/priv key pair has to be the same as the external one?
Or could I generate a self signed one for authentication and have a Letsencrypt Certificate in my Reverse Proxy for external access?

mgansler:
**Update:**

My Registry/Gitlab now use a self signed certificate internally. The LE-cert is loaded by HAProxy and handles SSL/TLS termination for external access. Everything's working (I can login/push to the registry w/o specifying the port).
The only "issue" I have right now is that Gitlab still includes the port in the `Container Registry` Port.
When I set `GITLAB_REGISTRY_PORT=""` and hardcode the port into the `gitlab-registry` nginx config file it still shows `registry.example.com:`. I think to get rid of the `:` changes must be made a Gitlab CE directly.
Push/Pull w/o the `:` works well though. So it's only a cosmetic issue.

Any Ideas on this?

SolidNerd:
@mgansler You could set `GITLAB_REGISTRY_PORT` to `443` this is the default port for https. So you should have the same behaviour.

SolidNerd:
@mgansler I'm quite not sure if we could leave registry keys in gitlab-ci.yml out. So if this works I could improve the templating of the registry part.

mgansler:
@SolidNerd I still have a `registry.crt` and `registry.key` which I pass to Gitlab/Registry. They are just different from the external SSL Certificate.
*BUT* what might work is generating these self signed when Gitlab starts. And if the `certs/` directory is shared between the Registry and Gitlab Container as in you example the user doesn't have to generate it manually.

sameersbn:
First of kudos to @SolidNerd for his excellent work on this PR. However at the moment I'm not very comfortable exposing the container registry integration as-is. The setup seems to be complicated and I would prefer holding back exposing the container registry options at least until we have a simple setup and cleanly documented instructions.

We have two options:

**Option 1**: release for 8.0.x versions that do not include any of the container registry integrations bits. So we have a proper release and everyone can upgrade to the latest versions while we continue working on this PR to clean it up, make it better and easy to setup.

**Option 2**: merge this PR and create RC releases instead.

Personally I prefer the first option. What do you guys thing?

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
I think Docker-gitlab is aim to install and use gitlab easy. So I prefer option1 too

nachokb:
@sameersbn given that upstream guys didn't even announce the registry until AFTER release, I would say there are many things still being decided, so I would also prefer option 1...

athieriot:
Where I am we need support for S3 storage anyway so I won't user Gitlab Registry just yet. Happy with both options if hopefully we see the feature released in the future :)

mgansler:
May I propose another option?
I just took a look at at the [Registry Image](https://github.com/docker/distribution-library-image) itself.
It looks really small and simple - two packages and a 24MB binary.
So why not include it here? Maybe in form of a Tag, but I don't think that will be necessary because of the small footprint and the registry will only be enabled and listening if the correct environment variables are set.
What do you think of this?

Anyway, I would be fine with either of the two options proposed by @sameersbn . Right now I'm running my container based on #705. And to be honest, the process isn't that complicated, once you figured out what to do, so if anything it's lacking a bit more documentation.

blandes:
@sameersbn I know that I'm not a common contributor but I think that the best route would be option 1. The thing is, the container registry is awesome and I'm so happy that they brought it into GitLab. 

But as a DevOps guy, that would mean that I would need to change my entire DevOps CI/CD cycles based on this registry. Like mentioned above, I want to be able to sync this to S3 just because of safety and so on. I would have to reconfigure every Developer's computer, or at least inform them, of the new migrations to the GitLab container registry. Let's pause on the registry inclusion and focus on the actual update. You guys are great at these images, and I know that you will make the right decisions but some of us need the new features that 8.8.0 - 8.8.2 bring now. It's not like I'm going to immediately switch my Dockerhub registry to this container registry without the proper safety measures anyways.

Regardless, thanks for all the hard work. You guys are the backbone.

bharrisau:
Seems like GitLab has separated the registry from the http proxy. You
should be able to disable the omnibus registry and point the url to another
docker container.

Looks like it is just the TLS setup between registry and gitlab that is
tricky.

SolidNerd:
Okay I will throw my 2 cents in how we could proceed. I understand the point that the container registry could lead to instability of this image. The current state is that the container registry is disabled by default. So any  user of this image should have no problems if he doesn't use this feature.

In the first point of view I like the idea from @mgansler to put in the binary in this image. This will make up a easier setup, but this produce a high coupling of dependencies. The first target should be GitLab by itself and not any services that are needed for a GitLab feature. So In my opinion the current way is the best. 

Most common problem of the container registry is the lack of documentation. It took me a day to find out that the official container documentation is not correct. Another problem of the container registry is that it's really new and have a lot of early adopting problems.

The best way how we could handle all this would be **option 1** upgrading to 8.8.X  and do not any configuration changes except they are implemented before and treat this PR as feature branch for the container registry.

How we could make a image based on this feature so that early adopters could test it and report problems because every knows that with the time you will be routine-blinded. Another step is to document how to setup the container registry. We could use my  [Gist](https://gist.github.com/SolidNerd/1bf47f85457e4f90eea541586f7290da) as starting point for the documentation.


@bharrisau It's the concept how the container registry works. To run it you need the docker registry for storing images and distribute images. The part of GitLab is to authenticate a user for doing this actions against the docker registry.

mgansler:
@SolidNerd regarding your [Gist](https://gist.github.com/SolidNerd/1bf47f85457e4f90eea541586f7290da):
I don't think you actually need the tls (certificate/key) entries in your config.
According to https://docs.docker.com/registry/configuration/#tls they are optional - and indeed, the Gitlab Nginx is handling SSL Termination for the registry. It all depends how you expose the registry to the outside world.
And you have three choices here, directly (`docker run -p 5000:5000 registry`), via the Gitlab Reverse Proxy (`docker run --net gitlab_default -p 172.x.y.1:5000:5000 registry`) or have another Reverse Proxy in front of Gitlab AND the registry.
In my opinion, the tls setting is only necessary with the direct option.

Also, if I understood it correctly, the rootcertbundle should (must?) not be the same as the SSL certificate.

Looking at https://github.com/docker/distribution/issues/520, I get the feeling that it should be indeed different from the SSL cert.

SolidNerd:
@mgansler  You are right. I didn't know that `tls` is optional this would make an integration a lot of easier. You could use the registry in the same `docker-compose.yml` for GitLab as a standalone services and configure all related stuff via environment variables.

To the `rootcertbundle` yes it's same like you said. So what i read from  [docker/distribution#520,](https://github.com/docker/distribution/issues/520) it can be every certificate but it's must be a trusted certificate or you must add to the trust store of os. 

Thanks for the clearification. I will wirite up a `docker-compose.yml` later to support everything.

mgansler:
@SolidNerd I don't think it even has to be trusted. It's just a private/public key used to sign/verify the token.
 I generated one myself and it works fine. I didn't add it to neither the registry nor the gitlab container.

I found these instructions somewhere (I can never remember how to create a certificate):
```bash
openssl req -new -newkey rsa:4096 > registry.csr
openssl rsa -in privkey.pem -out registry.key
openssl x509 -in registry.csr -out registry.crt -req -signkey registry.key -days 3650
```

vitorarins:
The docker registry is a real pain to install. Even a really good documented application like Portus is difficult to set it up, depending where you want to install things. So I think we are going to wait forever until we make things "easy" to set up. I say we could release this, after all it is optional, and work on the documentation. The gist is a good starting point, I can help showing what I did with HAProxy and systemd units for a CoreOS install of a registry. I started some work here: https://github.com/NeowayLabs/deploy-portus but there is a lot to do.

tkaefer:
Hi there - just my two Cents. 
Let's Encrypt is like no option for people that use GitLab with a corporate and protected environment. And there are actually on nginx-reverse-proxy container around that could do this trick, like https://github.com/eforce21/letsencrypt-nginx-proxy.
I would vote to just upgrade this image to GitLab 8.8.x and see if some one is able to do all of the certificate magic for own hosted CAs (which is mostly a pain somewhere in the lower area).

Nonetheless, I am really looking forward to see the docker registry included here - which might come later and I am totally fine with this.

Cheers.

lenovouser:
I would rather see you guys including the registry now so that people who really want to use it can do so - and work on refining the image and documentation later. I've been hogging this issue for the past days because I really want to use the registry :smile: 

Also, AFAIK is the whole certificate stuff for the registry unnecessary - since you proxy it with NGINX anyway like you do it with GitLab too.

ShakataGaNai:
Personally, I **really** want the registry integration. That was supremely exciting to me. However if getting the registry out the door "today" means a possibly not great setup that could potentially break people and/or break on upgrade, then I'd say wait on that portion. So Option 1 seems like the logical choice.

hanej:
In my view, the primary responsibility of GitLab, and by extension this Docker image, is to provide a stable environment for a centralized git repository.  Anything else is extra.  I would much rather have a reliable image of GitLab that works well than try to add something else in that may not be quite ready yet.

From what I've read, the GitLab Docker registry still needs some improvement as it requires Docker-in-Docker running in privileged mode.  That's not the way I want to operate.  As an administrator who needs GitLab to be reliable for the entire company's source code, I would much go with Option 1 and wait a little longer until the kinks get worked out.

mgansler:
@hanej what exactly has to run in privileged mode?
AFAIK, you MAY need privileged mode for certain CI functionality (spawning and linking services).
But this is nothing new.

Also: http://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/ (written by the guy behind dind)

I just tested it: for docker build/tag/rmi/push there is no need for privileged mode at all. All you need is the docker binary and access to the docker socket.

hanej:
@mgansler Your docker container has to run in privileged mode if you're using dind and GitLab runners

> Notice that it's using the privileged mode to start the build and service containers. If you want to use docker-in-docker mode, you always have to use privileged = true in your Docker containers.

http://docs.gitlab.com/ce/ci/docker/using_docker_build.html

hanej:
See also https://gitlab.com/gitlab-org/gitlab-ce/issues/17769

mgansler:
@hanej thanks for the link to the issue, but I don't completely understand their argument.
`docker build` (and therefor `docker run`) works just fine, `docker push` as well. So no limitations as far as I can tell.

Security is a different Story - mounting `/var/run/docker.sock` basically exposes the entire Host to users of gitlab - they could mount all kinds of directories. It could be used for spam and whatnot. But the same applies to dind. Only it's limited to everything inside this docker container.
But isn't it a good Idea to have the runners running on isolated hosts (VM?) anyway?

sameersbn:
I am going to cherry pick the commits from this PR and ready a release without the contaienr bits. This will be followed by the creation of a branch with the container specific stuff this PR add and will setup and automated build for it so that users who want to setup and use the container registry feature can do so.

Oh shit. just realized this PR makes one gaint commit. 😢

SolidNerd:
> At all. I can say that the container registry is working now. I guess that I have found a bug in GitLab that causes the issues with the container registry but i have fixed it locally. My next plan was to test it all and after that i will cleanup my commits and rebase it.

Sorry. I guess we should close this PR and make a new one for the upgrade only. I could do it if you want.

sameersbn:
@SolidNerd no worries.. I have the branch with your individual commits. Will be making the release a few minutes from now 😄

SolidNerd:
@sameersbn Do you upgrade directly to 8.8.2 ? And can you create the branch for the Container Registery so i can Start a new Prozent and working on it.

sameersbn:
I am publishing all the intermediate releases too. I will  create a branch with all your work and give you commit access to the repo 😉

sameersbn:
release 8.8.0 thru 8.8.2 have been published sans container registry support 😢

SolidNerd:
For everyone who wants to try the registry check this [#PR 708](https://github.com/sameersbn/docker-gitlab/pull/708) out.

sameersbn:
the builds are in queue.. may take a couple of minutes before you can pull them from dockerhub.

sameersbn:
@SolidNerd can you post the relavant links to the PR (like the gist your created
)

blandes:
Thank guys!

SolidNerd:
@sameersbn Yes. I see bug through the old commit and I guess it will currently not work. I have created a fix for that of with one my latest commits.

PR #708, opened 2016-06-01
Add Container Registry Support

sameersbn:
The container registry work from #697 (by @SolidNerd) has been migrated to this WIP PR (wip-container-registry branch) and will continue to be tested and refined until its ready to be merged into the master.

The `sameersbn/gitlab:latest-container-registry` is built from the `wip-container-registry` branch so users can test or start using the container registry feature in its current state.


SolidNerd:
I will post updates and changes and anything related to the container registry in this PR.

SolidNerd:
For everyone who wants to setup it i have been created a little [document](https://gist.github.com/SolidNerd/1bf47f85457e4f90eea541586f7290da) for that.

SolidNerd:
You need to be a little be patient. I found a bug currently this is caused by the old commit of mine. I will prepare a fix for it and you will see the change here. 

If you haven't pulled this tag you should wait or if you did it wait for the fix and pull it again.

tkaefer:
FTR: Thanks for keeping this registry integration alive. I really appreciate the work and the solution you have now chosen.

mgansler:
I just noticed an inconsistency of naming the env-vars in the README and the scripts used to configure gitlab. See #711

mgansler:
I discussed this with @SolidNerd quite a bit already but I guess we need all your opinions on this.
If you don't intend to use the built in Reverse Proxy for the Registry (e.g. you are using an external Reverse Proxy or your Registry is on another host) currently there are some issues.

Currently, `GITLAB_REGISTRY_PORT` defines, on which port the built in reverse proxy is listening as well as what Gitlab shows in the Container Registry Tab. This is perfect if you use a port different from `443`.

**Issue 1**:
This is purely cosmetic. If you are using port 443, you don't have to specify that port for `docker login` or `docker push/pull`

**Issue 2**:
Let's say you are using the built in Reverse Proxy and want it to listen on Port 5005.
But you have another Reverse Proxy in front of that, which listens in Port 5000. Then the information Gitlab shows you is actually wrong!
Using two Reverse Proxies may sound stupid at first, but when you setup Gitlab without SSL certificates it assumes you are running behind a Reverse Proxy - but the built in Reverse Proxy actually routes HTTP  Requests to the Gitlab Workhorse (see `assets/runtime/config/nginx/gitlab`)

I'm not sure what is the best way to resolve this right now, maybe I'll have an Idea tomorrow.

tkaefer:
@mgansler Thanks for bringing this up. I think the two-reverse-proxy setup is a valid example and not only a esoteric corner case (you already named SSL offloading). 

Issue 1 is the one that could be ignored in first place. 

Issue 2 leads me to the question if anybody is intending to have separated port mappings for the reverse proxy setups. I am not so sure how oft this would happen, since you could be quite free to use any port on the built-in reverse proxy and just align it with the one in front of it.
(Or maybe I did not get your concern right ;))

mgansler:
@tkaefer I guess my argument is, that these two settings can be independent and should therefor be configurable independently. I think the issue right now is, that you must define `GITLAB_REGISTRY_PORT`, otherwise the built in registry reverse proxy fails to start. But you don't necessarily want this to be the port to show up in the Gitlab UI.

Maybe the simplest solution would be to introduce something like `DISABLE_GITLAB_REGISTRY_REVERSE_PROXY`.

SolidNerd:
Yes indeed the current problem is that it depends on only a single variable. Yesterday i thought about it and I came to the same solultion like@mgansler.  It should be an seperate variable added that sets the nginx gitlab proxy port. But It needs a solution that you don't  need to configure both.

My proposal would be something like this:

```
$GITLAB_REGISTRY_PORT = ${GITLAB_REGISTRY_PORT :- } # Set Default registry to empty
if [[ -n $GITLAB_REGISTRY_PORT ]]; then
  NGINX_REGISTRY_PROXY_PORT = 443
else
  NGINX_REGISTRY_PROXY_PORT = $GITLAB_REGISTRY_PORT
fi
```

This sould solve both issues and extend the configuration of the container

mgansler:
@SolidNerd I agree, `GITLAB_REGISTRY_PORT` must not have a default value, otherwise it's impossible to set to empty.

SolidNerd:
I updated this branch to make it even with master. So the current GitLab Version is `8.8.3` then.

lenovouser:
@SolidNerd any progress?

SolidNerd:
@lenovouser  Yes indeed. Backup can be restored now and so i guess it will be usable for master. I will push all changes in the evening so you can try it out.

SolidNerd:
Some last big issue i found is when you run your own registry under the same domain. The proxy will not be triggered and every request will be directed to gitlab. So should we ignore this i don't find any related issue on GitLab.com.

umputun:
@SolidNerd by "the same domain" you meant FQDN? If this is the case and someone running private registry on git.example.com and also using the same git.example.com for gitlab - I'd say ignore this issue. In my opinion if one hosting both registry and gitlab on the same box he probably will want to leverage gitlab's registry anyway.

SolidNerd:
@umputun Yes I meant FQDN. It makes sense to ignore this. So I will start to write the documentation for the integration today.

cvle:
Thanks! I got it setup using your instructions and in my case, it was straight forward!

SolidNerd:
8.8.4 is now available on this branch.

boonkerz:
Wow thx for made this possible it works and can replace quay.io :)

boonkerz:
so all settings are made i can push pull directly to my gitlab instance.
the only thing which not work ist the registry display in the project.
nothing is display. 

the registry log says
`2016-06-16 09:36:25,312 INFO: 172.17.0.16 - - [16/Jun/2016:09:36:25] "GET /v2/tp/test-old-docker/tags/list HTTP/1.1" 404 233 "-" "Faraday v0.9.2`

but all other works

SolidNerd:
@boonkerz Thanks for your reporting. I guess your problem is that `GITLAB_REGISTRY_API_URL` is wrong but in this case push/pull should not work as well. Futhermore you using the nginx in gitlab for proxying the registry (This is the case if you using using the url that is shown in GitLab). On my testing instance everything works as excepted.

boonkerz:
@SolidNerd it looks like registry does not support v2

> root@108122ce8fdf:/#  curl -X GET http://127.0.0.1:5000/v2/   
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>404 Not Found</title>
<h1>Not Found</h1>
<p>The requested URL was not found on the server.  If you entered the URL manually please check your spelling and try again.</p>
root@108122ce8fdf:/#  curl -X GET http://127.0.0.1:5000/v1/_ping
true
root@108122ce8fdf:/#

lenovouser:
@SolidNerd any ETA on when this will be released with an option to upgrade existing installations with it?

SolidNerd:
@lenovouser  There are some trifles on my list that need to be done. I will assume that I'm finished by no later than saturday. This includes to upgrade an existing installation.

SolidNerd:
I'm finished this feature is stable for me now. If you want to adopt before the merge is passed use the tag.

@sameersbn I added a seperate doc folder for the documentation of this feature. Can I merge or is something left with this feature.

Ilya-Kuchaev:
Any chance to see it merged? :)

mgansler:
@SolidNerd Please check #741

SolidNerd:
At all I will prepare a new release please use then `8.8.5-1`.

sameersbn:
Thanks for the great work you guys! I'm sorry I wasn't involved with this PR, but I trust @SolidNerd's judgement on this matter 😉

sameersbn:
p.s. I'm taking down the `latest-container-registry` builds from dockerhub.

toch:
thanks a lot @SolidNerd for that great work. I successfully deployed it yesterday (option 1 of your gisted doc).

PR #711, opened 2016-06-01
GITLAB_REGISTRY_PATH is really GITLAB_REGISTRY_DIR

mgansler:
The README says `GITLAB_REGISTRY_PATH`, but in all scripts `GITLAB_REGISTRY_DIR` is used.


SolidNerd:
Thanks for the pull request. You are right but I guess my implementation is wrong. Have a look on the [official documentation](https://gitlab.com/gitlab-org/gitlab-ce/blob/8-8-stable/doc/administration/container_registry.md) it is called `path`. I will merge it  and refactor everything easily later.

PR #714, opened 2016-06-02
disable registry reverse proxy

mgansler:
If the user wants to run his own Reverse Proxy anyway there is no need to enable the built in one.


SolidNerd:
I don't like the name of the env variable. It could misleads someone who doesn't know that it's use to change stuff on NGINX and not GitLab. Another problem of the naming is variable is negative and the default value aswell. This could outface a user.

I would prefer this `NGINX_GITLAB_REGISTRY_ENABLED`. This speficies thats a nginx related variable and it's default enab

@sameersbn what do you think about it ?

SolidNerd:
My current implementation will now work like the normal ssl proxy for gitlab if you didn't provide a cert and key for `SSL_REGISTRY_CERT` & `SSL_REGISTRY_KEY`. It will be assumed that you are running the registry on your reverse proxy. So In this case an additional environment variable isn't required.

PR #715, opened 2016-06-02
default value for GITLAB_REGISTRY_PORT is 443

mgansler:


assets/runtime/config/gitlabhq/gitlab.yml
```diff
@@ -178,13 +178,13 @@ production: &base
...
-    # key_path: config/registry.key
-    # path: shared/registry
-    # issuer: gitlab-issuer
+    enabled: {{GITLAB_REGISTRY_ENABLED}}    # Enables GitLab Container Registry Defaults to `true`
+    host: {{GITLAB_REGISTRY_HOST}}          #registry.example.com
+    port: {{GITLAB_REGISTRY_PORT}}          # 5005
+    api_url: {{GITLAB_REGISTRY_API_URL}}    #http://localhost:5000/ # internal address to the registry, will be used by GitLab to directly communicate with API
+    key: {{GITLAB_REGISTRY_KEY_PATH}}       # config/registry.key
+    path: {{GITLAB_REGISTRY_DIR}}
+    issuer: {{GITLAB_REGISTRY_ISSUER}}       #gitlab-issuer
```
IlyaSemenov: Inconsistent comment indentation (also inconsistent spaces after # here and above)

thynson:
+1

SolidNerd:
Could you rebase it. So i can merge your PR ?

mgansler:
I'm not sure I can. There is no default for `GITLAB_REGISTRY_PORT ` right now in your branch, right?
And it may be the Rum talking, but I think their might be an issue - if `GITLAB_REGISTRY_PORT` is not set, `NGINX_REGISTRY_PROXY_PORT` is set to `443`. But if certificates for Gitlab and the Registry are provided, the current setup tries to start up two instances of Nginx listening on Port 443 which won't work.

SolidNerd:
Nope 😄 .  There will be only one instance of NGINX started all configuration related to gitlab and the registry are in [server blocks](https://www.nginx.com/resources/wiki/start/topics/examples/server_blocks/) (Apache = Virtual Hosts). So NGINX will handle this for us. Now it would be nice to change the title to something that is related with this commit.

SolidNerd:
Yeah this make a lot of more sense to me. Thank's your hard work.

PR #741, opened 2016-06-21
Wip container registry fixes

mgansler:
* `GITLAB_REGISTRY_PORT` is never empty, so we can just assign it's value to `NGINX_REGISTRY_PROXY_PORT`
* `GITLAB_REGISTRY_PORT` should not be overwritten globally. If someone changes the order of functions their might be side effects then.

Could someone please verify if this works? I don't have a docker host available at this time, I could only test the `local` thing with this test:
```bash
#!/bin/bash

FOO=BAR

test() {
	echo 1 ${FOO}
	local FOO=FOO
	echo 2 ${FOO}
}

echo 3 ${FOO}
test
echo 4 ${FOO}
```

```bash
$ ./test.sh
3 BAR
1 BAR
2 FOO
4 BAR
```


mgansler:
Closing because target was already merged.

PR #743, opened 2016-06-21
Fix presentation of SSL_CIPHERS

SolidNerd:
This should be the same like in https://github.com/sameersbn/docker-gitlab/blob/master/assets/runtime/env-defaults#L133 . Otherwise it could leads into problems when you try to configure it and it doesn't work  because of the bangs (!).

/cc @splattael .


splattael:
@SolidNerd I thought about quoting the default example in README as well but didn't do on purpose.

Reasons:
When copied+pasted *with* quotes
* It will break in `docker-compose.yml` (because of YAML)
* It will break in a envfile (`--env-file`)

No other default value in the README is actually quoted. The values are quoted in longer `docker run` examples.

We could add a generic hint that about quoting values when using `docker run -e`.

WDYT?

SolidNerd:
Okay thanks your reply I forget the cases with `docker-compose` and `--env-file`. For me it's the best as it is now. So I will close this.

PR #745, opened 2016-06-21
Fix typo in container_registry.md

ToothlessGear:
Replace "enviroment" with "environment"


sameersbn:
ouch! @SolidNerd does the container bits need more testing?

SolidNerd:
Oh thanks for you report. The origin of this is issue is in my Gist. I copied it from there. 

@sameersbn Why not. It's stable for me and I hope that i found every evidently bug. I tested every setup that was in my imagination. 😇

SolidNerd:
Thanks. I will merge this now.

PR #746, opened 2016-06-21
remove unneeded NGINX_REGISTRY_PROXY_PORT variable

mgansler:
* `GITLAB_REGISTRY_PORT` is never empty, so we can just assign it's value to `NGINX_REGISTRY_PROXY_PORT`
* `GITLAB_REGISTRY_PORT` should not be overwritten globally. If someone changes the order of functions their might be side effects then.

Could someone please verify if this works? I don't have a docker host available at this time, I could only test the `local` thing with this test:
```bash
#!/bin/bash

FOO=BAR

test() {
	echo 1 ${FOO}
	local FOO=FOO
	echo 2 ${FOO}
}

echo 3 ${FOO}
test
echo 4 ${FOO}
```

```bash
$ ./test.sh
3 BAR
1 BAR
2 FOO
4 BAR
```


mgansler:
@sameersbn @SolidNerd ping! Any opinions on this?

sameersbn:
@SolidNerd can you please take a look at this

SolidNerd:
You are right overwrite the global `GITLAB_REGISTRY_PORT` is not a good idea. I test you example and the idea  of local variable is good. This changes the scope of variable but this is what we want. In my point of view scoping `GITLAB_REGISTRY_PORT` to local resolve the problem of set an an empty value. So we could set `GITLAB_REGISTRY_PORT` only empty during the configuration of gitlab registry.

Here my thoughts about it how we could realize it. `NGINX_REGISTRY_PROXY_PORT` should be **removed** than.

```bash
gitlab_configure_registry(){
  echo "Configuring gitlab::registry..."

  if [[ $GITLAB_REGISTRY_PORT == 443 ]]; then
    # Sets GITLAB_REGISTRY_PORT empty during this function.
    # This helps us to add an empty key to `.gitlab-ci.yml` .
    # It needs to be empty because 443 is the default https port and doesn't need a mention
    # so it will not be displayed on the ui.
    local GITLAB_REGISTRY_PORT
  fi

  update_template ${GITLAB_CONFIG} \
    GITLAB_REGISTRY_ENABLED \
    GITLAB_REGISTRY_DIR \
    GITLAB_REGISTRY_HOST \
    GITLAB_REGISTRY_PORT \
    GITLAB_REGISTRY_API_URL \
    GITLAB_REGISTRY_KEY_PATH \
    GITLAB_REGISTRY_ISSUER

}


nginx_configure_gitlab_registry() {
  if [[ $GITLAB_REGISTRY_ENABLED == true && -f ${SSL_REGISTRY_CERT_PATH} && -f ${SSL_REGISTRY_KEY_PATH} ]]; then
    echo "Configuring nginx::gitlab-registry..."
    update_template ${GITLAB_REGISTRY_NGINX_CONFIG} \
      GITLAB_LOG_DIR \
      GITLAB_REGISTRY_PORT \
      GITLAB_REGISTRY_HOST \
      GITLAB_REGISTRY_API_URL \
      SSL_REGISTRY_KEY_PATH \
      SSL_REGISTRY_CERT_PATH
  fi
}

```

mgansler:
Testing my changes now, will remove `WIP` from the title when I'm done.

SolidNerd:
I found a strange behaviour of my propose. If I set `GITLAB_REGISTRY_PORT` to `443` registry_port will be set to `443` as well. This is not what we want. But when I didn't set GITLAB_REGISTRY_PORT it will be empty.

mgansler:
Same here, that's really strange.

mgansler:
Well, there is an easy fix - explicitly set `GITLAB_REGISTRY_PORT` to empty string: `local GITLAB_REGISTRY_PORT=""`

SolidNerd:
Yep that's correct this reminds me on the problem in the past how I solved it :) .

assets/runtime/functions
```diff
@@ -817,26 +817,24 @@ gitlab_configure_project_features() {
...
-    update_template ${GITLAB_CONFIG} \
-      GITLAB_REGISTRY_ENABLED \
-      GITLAB_REGISTRY_DIR \
-      GITLAB_REGISTRY_HOST \
-      GITLAB_REGISTRY_PORT \
-      GITLAB_REGISTRY_API_URL \
-      GITLAB_REGISTRY_KEY_PATH \
-      GITLAB_REGISTRY_ISSUER
+
+  if [[ $GITLAB_REGISTRY_PORT == 443 ]]; then
```
SolidNerd: Could we change `$GITLAB_REGISTRY_PORT` to `${GITLAB_REGISTRY_PORT}` to be more consistent. I know this is an error of my own.

mgansler: done :)

SolidNerd:
Thanks for your hard work and patient on this PR. I will merge this now, because it will provide us more stabililty as my first approach. 

@sameersbn  Should I  make a new release with `8.9.1-1` to add a this improvement as an seperate update or should it wait for a next update and it will be in latest only.

sameersbn:
@SolidNerd If this is a desperately required fix then yes, you can make a hotfix release. In other cases we can wait for a few days for the next point release by gitlab.

SolidNerd:
Okay I will merge it and we wait for the next gitlab release

PR #759, opened 2016-06-24
Fix #757 | Email notifications stopped working after upgrade to 8.9

### LINKED ISSUE #757: Email notifications stopped working after upgrade to 8.9
Error is also reported on the following page: `/admin/health_check`

> Current Status:  Unhealthy
> no sendmail executable found.

ntrifunovic:
Fix for issue #757.

Added `exim4-daemon-light` package.


SolidNerd:
I will close this PR because it's isn't a fix for issue #757. Yes It makes the heath_check green again but mails won't be sent.  I tested my PR #763  and health_check goes green again and mails will be sent. The origin of the problem comes from upstream like I said in #757 .

PR #791, opened 2016-07-13
WIP Seperate NGINX in a container

solidnerd:
Hey,
I started to working on a seperation of NGINX because GitLab has now `gitlab-workhorse` to handle their http communication. I want to discuss this here if we  want to remove it. It has pros and cons and we should think about them.  My current state is that nginx is removed from this image and workhorse handles it. I tested it and it worked for me.


SolidNerd:
ping @sameersbn WDYT ?

sameersbn:
@SolidNerd I've avoided this change because removal of NGINX means that we can no longer have SSL termination in the image and users would always have to use a load balancer to provide the SSL termination.

I'm sure there are at least a few users using SSL termination this way.  I wouldn't want to break this for them.

What are your thoughts?

kiview:
@sameersbn I think your point considering backwards compatibility is valid, however, aren't most users already using a tailored version of the `docker-compose.yml` file? You could add the NGINX there (and hope users uprading to a newer version will notice this change).

lenovouser:
I am actually for removing it from the image as I proxy it with another container too. I also think that most people do the same.

SolidNerd:
My intention was to be able to do automatic ssl  acquisition. In the current state it's not easy to implement with a separate container it's a lot of easier. I had the same consideration about SSL and this image. Currently I think my approach is not the best way how we can solve this.
A better solution would be in my opinion now to have the ability to enable nginx or not and let workhorse work under tcp(8080) instead of a socket. This give us the ability that the user can choose what he wants. It's not a complete clean way but this should unite a user.  This is why I want to discuss.

sameersbn:
I am okay with the removal of NGINX from the image. But it should be done gradually in phases to provide users time to migrate. I am thinking anywhere from 3 to 6 months for the migration to be completed. 

There are a number of institutions using this image in their production environments and it'll be wrong to yank the nginx without any warnings. Here's my proposal to carrying out this change:

#### Phase 1 (duration: 2 month) 
- Add `NGINX_ENABLED` env variable that defaults to `true`
- Test internally over a couple of gitlab releases to ensure that everything is going to be fine.

#### Phase 2 (duration: 1 months)
- Provide the necessary samples/templates for the migration
- Announce to users in whatever ways possible
  + Mark sections of README (SSL config, etc) as deprecated
  + Add warnings to the container output
  + Update the Announcements issue

#### Phase 3 (duration: 3 months)
- Set `NGINX_ENABLED=false` by default

#### Phase 4
- Completely remove NGINX and the `NGINX_ENABLED` flags

What do you think?

SolidNerd:
@sameersbn I don't want to force you into a corner. I think a time frame of removing NGINX is great. To give the people a migration window. I'm fine with your proposal.

I think we could serve both a container without NGINX and a container with it. This requires a little restructure but It should work.

This could be done if we use different `FROM`. An example would be like this:

GitLab Image without NGINX .

```
FROM sameersbn/ubuntu:14.04.20160608

<GITLAB_PART>

```

And then we create an Image with extension of nginx:

```
FROM sameersbn/gitlab:8.9.6

<NGINX_PART>

```

They need separate tags so we could have :
1. `sameersbn/gitlab:8.9.6` 
2. `sameersbn/gitlab:8.9.6-nginx` 

The complete separation could be done through different folder so it's simpler to maintain it. This proposal should be done in a migration phase as well.

sameersbn:
> They need separate tags so we could have :
> 
>     sameersbn/gitlab:8.9.6
>     sameersbn/gitlab:8.9.6-nginx

IMO `sameersbn/gitlab:8.9.6` should always be the main version that everyone is expected to use. If you think we should have a separate version without nginx, then it's  the one that should be named differently.

Anyways, I am of the opinion that adding a `NGINX_ENABLED` env variable should be good enough. I can't  think of any maintenance issues during the 6 month transition period. It would also allow us to rollback this change, without anyone ever noticing, in case it does not work out for certain cases.

SolidNerd:
Okay. I will start to work on this now, but I will do this on a separate PR.

f3l1x:
I vote for drop nginx in 4 phases, it look's great.

frank-dspeed:
Separation of Nginx is a logic Consequence as we separated all other stuff also!

- I am for the -nginx naming and not for the -standalone naming but i would be also fine with -standalone
or
- -workhorse i think since services like CloudFlare exist the standard is that people load balance that only the once running on local machines would use internal nginx.

as also Omnibus Dockerfile exists and that for those cases we should stay at what we do best building from source separated containers for server deployments on real servers.

PR #793, opened 2016-07-17
Use tcp for workhorse communication.

SolidNerd:
Hey, 
this is one part of #791 migration. I would like to use this to make it simpler to change a load balancer or use no one for GitLab.  We could know adopt a load balancer container in the gitlab network or create a new network with the gitlab container and every other container can communicate with gitlab over `8181`. This doesn't impact the current behaviour it adds only more flexibility. WDYT ? @sameersbn


SolidNerd:
ping @sameersbn

PR #794, opened 2016-07-17
Add NGINX_ENABLED environment variable

solidnerd:
This is the second part of #791. This introduces a new variable to configure if you want NGINX as a load balancer for ssl or other stuff. If you don't want this you can disable it but it's **EXPERIMENTAL**. You have to use than port `8181` to speak plain http this will be able when #793  got merged. After this we should merge this PR.  We have migration phases for this have a look at the phases below.
#### Phase 1 (duration: 2 month)
- Add `NGINX_ENABLED` env variable that defaults to `true`
- Test internally over a couple of gitlab releases to ensure that everything is going to be fine.
#### Phase 2 (duration: 1 months)
- Provide the necessary samples/templates for the migration
- Announce to users in whatever ways possible
  - Mark sections of README (SSL config, etc) as deprecated
  - Add warnings to the container output
  - Update the Announcements issue
#### Phase 3 (duration: 3 months)
- Set `NGINX_ENABLED=false` by default
#### Phase 4
- Completely remove NGINX and the `NGINX_ENABLED` flags

At least we should make an Announcement of the removal of NGINX with all phases.


SolidNerd:
ping @sameersbn

sameersbn:
Hey @SolidNerd, Sorry about the delay.
I think we should warn users trying to install SSL certificates when they set `NGINX_ENABLED=false` cause its definitely not going to fly.

lenovouser:
Do you want to release this with the `8.10` release coming today, or release that one first and then add this?

SolidNerd:
I don't think that we release this `8.10`. This could be a separate merge. If you don't want to use nginx use workhorse directly with the `8181` port. Have a look at https://github.com/sameersbn/docker-gitlab/commit/70dc29de1a1c6b226c4e5f9e8c45fb3ae0136170 . If you get further questions to this open an issue for that.

SolidNerd:
@sameersbn I will prepare a separate document for this case and describe the pros and cons of it.

sameersbn:
@SolidNerd I was thinking maybe we should revert 70dc29de1a1c6b226c4e5f9e8c45fb3ae0136170. In this PR we can configure `gitlab-workhorse` to listen on port `80` when `NGINX_ENABLED=true` using `sed`. WDYT?

SolidNerd:
@sameersbn  I don't like the idea of reverting https://github.com/sameersbn/docker-gitlab/commit/70dc29de1a1c6b226c4e5f9e8c45fb3ae0136170 . But I assume that you meaning to set gitlab-workhorse to `80` when `NGINX_ENABLED=false`. Yeah we could use sed for it.

SolidNerd:
If we run workhorse on `80`. The process must be run by root, because linux wants to run ports <= 1024 must be run by root. The problem of workhorse is that it's currently not capable to drop the privileges.

rugdealer:
Wondering if this is still being integrated because I would love to use the [nginx-proxy container](https://github.com/jwilder/nginx-proxy) with the gitlab container.

lenovouser:
I second @rugdealer

lorenz:
I'd like this (and the other associated PR) to be merged since I'm using a different frontend for SSL termination already which makes Nginx inside this image useless and actually less efficient and causes Unicorn to see the wrong external IPs (since the Nginx in this image does not itself use X-Forwarded-For, which is fine if it is the frontend, but not if it's behind another proxy).

PR #799, opened 2016-07-23
Fix nginx Host header with Port to receive the correct redirects

SolidNerd:
Like i said in https://github.com/sameersbn/docker-gitlab/commit/48cf19f9571cc5519908ee2d5ae0df71a9e5871e . This will set the port as well to create the correct redirects if you not using the default Protocol Ports like `80` and `443`.


sameersbn:
@SolidNerd Did you test it?

SolidNerd:
Yep.

Am 23.07.2016 6:12 nachm. schrieb "Sameer Naik" <notifications@github.com>:

PR #808, opened 2016-07-26
Google OAuth - Handle multiple restricted domains

achretien:
Since https://gitlab.com/gitlab-org/gitlab-ce/commit/85cb5635babe5aa2c051fa2f40be4b28a106640f the google oauth handle multiple hosted domain validation.

Now we can use a list of restricted domains.


SolidNerd:
Hey, 
thanks for your  work. Could you rebase your commits to a single commit. And did you tested it ?

achretien:
Hi, I just squash the two commits.
And yes I m running my fork in production

SolidNerd:
👍 . One last thing is the commit message I would prefer something like `Improve OAUTH_GOOGLE_RESTRICT_DOMAIN to have mutltiple restricted domains`. This should help us to search something in the history instead of your title. Because In the future we won't remind us of GITLAB URL but the env key and we can have a look up for it in `env-defaults` or `functions` .

achretien:
Everything should be ok. Thanks for all the work already done on this repo 👍 .

PR #833, opened 2016-08-07
Add microbadger layers badge.

SolidNerd:
This is an replacement for the imagelayers.io badge.


SolidNerd:
@sameersbn What do you think ? If you okay with that than feel free to merge it. Accordingly to this we should think about #738 . It's for the same site.

PR #839, opened 2016-08-12
Grammar patch

nsymms:


nsymms:
Crap. Hit the wrong button. Sorry.

PR #840, opened 2016-08-13
Update docker-compose.yml example

ngpestelos:
1. Adds a `registry` dependency to the `gitlab` service

`gitlab` refers to the `registry` service through Docker's internal
network.
1. Specify HTTPS configuration variables

This example shows which parameters are needed to enable HTTPS in
GitLab. Also added a port mapping for HTTPS.

### DIFF
--- a/docs/container_registry.md
+++ b/docs/container_registry.md
@@ -83,6 +83,7 @@ services:
     - --loglevel warning
     volumes:
     - ./redis:/var/lib/redis:Z
+
   postgresql:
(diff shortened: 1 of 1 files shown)


SolidNerd:
Since you changed the configuration to https. I will be required to have the certificate of the gitlab container in the registry container when it's self signed or the ca not exists. So in this case `REGISTRY_AUTH_TOKEN_REALM` must be changed as well.

PR #842, opened 2016-08-17
Change `GITLAB_CLONE_URL` from github to gitlab directly

solidnerd:
So I would prefer to use GitLab directly as clone url. I saw the first difference now on GitHub (8.11.0-rc2 isn't released). All other realated components will be fetched from GiLlab. I think it would be more consistent. If you like it feel free to merge it otherwise I would like to know what are you thinking about it.

### DIFF
--- a/assets/build/install.sh
+++ b/assets/build/install.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-GITLAB_CLONE_URL=https://github.com/gitlabhq/gitlabhq.git
+GITLAB_CLONE_URL=https://gitlab.com/gitlab-org/gitlab-ce.git
 GITLAB_SHELL_URL=https://gitlab.com/gitlab-org/gitlab-shell/repository/archive.tar.gz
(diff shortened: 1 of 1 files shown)


SolidNerd:
ping @sameersbn

sameersbn:
I've switched back and forth between github.com and gitlab.com clone urls on multiple occasions in the past. The last time I switched it was when the image builds were failing during to issues to perform a git shallow clone from gitlab.com.

On that particular occasion no immediate fixes were being proposed and out of frustration i moved it back to github.com. Ideally I would perfer to clone all repos from github.com, but some of the projects are not mirrored on github.com.

SolidNerd:
I understand your frustration. In this case it's okay for me that we let `GITLAB_CLONE_URL` pointing to GitHub.

solidnerd:
Hey @sameersbn,

I got this problem again that tag does not exist on GitHub. I had this case with `8.14.2` (#994) and now with `8.15.0`. I will switch to the gitlab clone url so it will work i tested it before. This should help us to make our releases easier as before. If we get problems with GitLab by shallow cloning we could switch back again and ping sometimes somebody that updates the tag on github.

PR #847, opened 2016-08-19
GitLab 8.10.7

SolidNerd:
Hey,
I prepared everything for the new release. I test it and remove later my WIP prefix when it's ready to merge.


SolidNerd:
I got a problem during the upgrade process. The problem is that when you upgrade your container to `8.10.7` the login form is missing. So that need to be resolved.

SolidNerd:
I resolved it by not jumping directly from `8.10.4` to `8.10.7`. I upgraded all pervious versions.

SolidNerd:
I wll start to :shipit:

PR #852, opened 2016-08-22
GitLab 8.11.0

SolidNerd:
Hey,

I prepared everything for 8.11.0 we should test it now after this I will start with the release.


Dockerfile
```diff
@@ -1,9 +1,10 @@
 FROM sameersbn/ubuntu:14.04.20160817
 MAINTAINER sameer@damagehead.com
 
-ENV GITLAB_VERSION=8.10.7 \
+ENV GITLAB_VERSION=8.11.0 \
+    RUBY_VERSION=2.3 \
```
splattael: YES! Finally :-)

sameersbn:
@SolidNerd The `config/resque.yml` format seems to have changed

SolidNerd:
@sameersbn You are right. But this will enable redis sentinel as well. Do we support it with it, because GitLab says that is a new experimental feature in `8.11.0`. I will update it and test it how sentinel works. For more informations have look at  http://docs.gitlab.com/ce/administration/high_availability/redis.html#experimental-redis-sentinel-support



Another step that needs to be done is  to create the following keys in `secrets.yml`. There are new keys.

```yml
secret_key_base: 
otp_key_base: 
```
If we don't do that we got the following message during container startup.

> gitlab_1      | Missing Rails.application.secrets.secret_key_base for production environment. The secret will be generated and stored in config/secrets.yml.
> gitlab_1      | Missing Rails.application.secrets.otp_key_base for production environment. The secret will be generated and stored in config/secrets.yml.

I will do the required steps today in the evening.

SolidNerd:
And we should support Koding. This needs to be tested as well.

sameersbn:
>  this will enable redis sentinel as well. Do we support it ?
No. At least not in this release. But i noticed that the config file has changed for single redis instance

Used to be:

```yaml
production: unix:/var/run/redis/redis.sock
```

Now it is: 

```yaml
production:
  url: unix:/var/run/redis/redis.sock
```

> So we need to introduce 2 two variables for the keys

Perfectly fine add these variables.

> I will do the required steps today in the evening.

Thanks for your amazing efforts.

invader35:
I currently test this on my server, and till now it seems to work nicely.

SolidNerd:
Im finished with testing. I will start to release it.

PR #856, opened 2016-08-23
better rake command with --user instead

pierreozoux:
Use the `--user` argument instead of `sudo`


solidnerd:
Sorry for my late late response. This is a nice command 👍 .

sameersbn:
Sorry, I didn't comment on this PR earlier. But there are multiple instances in the readme that we could update for this change.

PR #860, opened 2016-08-24
Explain 8.11's secrets changes in more detail

smcgivern:
@SolidNerd does this look OK? I realise there's some repetition, but I wanted to help out people who only read the Quick Start :wink:


SolidNerd:
Thanks 👍 . I will update now the section of upgrading after this.

PR #882, opened 2016-08-31
UID/GID non-unique fix

GJRTimmer:
Complete details for PR: https://github.com/sameersbn/docker-postgresql/pull/78

allow non-unique id's, because the gid selected by the user may already be in use within the container. This fix solves that issue.

Docker error log:
```
Adapting uid and gid for postgres:postgres to 1024:101
groupmod: GID '101' already exists
```

Error within log means that GID: 101 is already in use within the container.
By using groupmod with option ```-o``` it will be allowed.


solidnerd:
👍

GJRTimmer:
Thanks, II hope you have time for the same Pull request for the postgresql.
Keep up the excellent work.

sameersbn:
@GJRTimmer just did.

PR #907, opened 2016-09-28
Allow for a different directory to store ssh host keys.

tbeadle:
Add a variable called SSH_HOST_KEYS_DIR that can be set to a directory that
will be used as the directory where ssh host keys are stored.  It defaults to
${GITLAB_DATA_DIR}/ssh.


assets/runtime/env-defaults
```diff
@@ -24,6 +24,9 @@ fi
 GITLAB_SSH_HOST=${GITLAB_SSH_HOST:-$GITLAB_HOST}
 GITLAB_SSH_PORT=${GITLAB_SSH_PORT:-$GITLAB_SHELL_SSH_PORT} # for backwards compatibility
 GITLAB_SSH_PORT=${GITLAB_SSH_PORT:-22}
+SSH_HOST_KEYS_DIR=${SSH_HOST_KEYS_DIR:-${GITLAB_DATA_DIR}/ssh/}
+SSH_HOST_KEYS_DIR=${SSH_HOST_KEYS_DIR%/}/
```
solidnerd: Duplication of line 27. I think you should write line 27 like `SSH_HOST_KEYS_DIR=${SSH_HOST_KEYS_DIR:-${GITLAB_DATA_DIR}/ssh}` and the removal of the trailing backslash in line 28 is obsolete in my opinion.  Because you remove it and add the slash again this is a little bit confusing 😅 .

solidnerd: I would write like `SSH_HOST_KEYS_DIR=${SSH_HOST_KEYS_DIR:-${GITLAB_DATA_DIR}/ssh}` this more common to the other directories.

solidnerd: In line 28 it's a little bit confusing 😅 . Why do you remove first the trailing backslash and add it again.

tbeadle: The reason I did it like that was to allow the user to define the variable with or without a trailing slash.  Then, after these statements, it would be sure to only end in one slash.

PR #909, opened 2016-09-29
Add missing SMTP_SSL to smtp_settings.rb

gianmarcotoso:
`SMTP_SSL` was missing from smtp_settings.rb, making it impossible to keep an SSL configuration (one had to manually edit the file every time the container started up). This should solve the problem.

### DIFF
--- a/assets/runtime/config/gitlabhq/smtp_settings.rb
+++ b/assets/runtime/config/gitlabhq/smtp_settings.rb
@@ -23,6 +23,6 @@ if Rails.env.production?
     ca_path: "{{SMTP_CA_PATH}}",
     ca_file: "{{SMTP_CA_FILE}}",
     tls: {{SMTP_TLS}},
-	ssl: {{SMTP_SSL}}
+    ssl: {{SMTP_SSL}}
   }
 end
--- a/assets/runtime/config/gitlabhq/smtp_settings.rb
+++ b/assets/runtime/config/gitlabhq/smtp_settings.rb
@@ -22,6 +22,7 @@ if Rails.env.production?
     openssl_verify_mode: "{{SMTP_OPENSSL_VERIFY_MODE}}",
     ca_path: "{{SMTP_CA_PATH}}",
     ca_file: "{{SMTP_CA_FILE}}",
-    tls: {{SMTP_TLS}}
+    tls: {{SMTP_TLS}},
+	ssl: {{SMTP_SSL}}
   }
 end


solidnerd:
Hey, 
why didn't work it when you set `SMTP_TLS` to `true`. Does your SMTP server negotiates only SSLv3 or lower ?

gianmarcotoso:
I honestly don't know, I was trying with Mailgun and it didn't connect on 587 TLS, while it works with both SSL on 465 and STARTTLS over port 25. Regardless of my specific use case, though, the setting is valid and it should be included.

solidnerd:
I'm quite not sure if you PR solves your problem for you and everyone else. For me it seems that you configure your mail settings not correctly. I had a look at possible omnibus configurations and here is an example for [Mailgun](https://docs.gitlab.com/omnibus/settings/smtp.html#mailgun). The equivalent configuration for our docker image would be:

```
SMTP_HOST=smtp.mailgun.org
SMTP_PORT=587
SMTP_AUTHENTICATION=plain
SMTP_STARTTLS=true
SMTP_TLS=false
SMTP_USER=postmaster@mg.gitlab.com
SMTP_PASS= 8b6ffrmle180
SMTP_DOMAIN=mg.gitlab.com
```

Could you check it if it works like that ?

gianmarcotoso:
I think you misunderstood the intention of my PR: it's not about resolving
a problem with my configuration (I am happily mailing away with TLS), it's
about adding a valid and recognized setting inside the configuration. If
some chap has an SSL only mailserver he won't be able to use this repo
unless he goes and modifies by hand the file every time the container
restarts. This PR solves that, nothing more and nothing less :)

solidnerd:
Okay you are right I misunderstood your intention. But a problem what I see is what happens when a user use `SMTP_SSL=true` and `SMTP_TLS=true`. Is it possible to run SSL and TLS on the same port ? And how handles ActionMailer this ?

gianmarcotoso:
Well that would be an issue outside of the scope of your repository: it's
possible to make all sorts of weird configuration errors, if they need to
be validated it's up to GitLab, not Docker. ;)

sameersbn:
> some chap has an SSL only mailserver he won't be able to use this repo
I think in this case you need to set:

```
SMTP_TLS=true
SMTP_STARTTLS=false
```

sameersbn:
FYI, I do not think this PR should be merged if my last suggestion works (it most probably will)

gianmarcotoso:
It does not, at least it did not in my configuration. What did work, though, was setting the `ssl` option within the configuration file to `true`. I don't really see what the problem is with this: the parameter is _valid_, but it's not carried over if set as an environment variable.

solidnerd:
I agree with @sameersbn . SSL and TLS are two different protocols. I can't predict the impact of this change. I can't create a tangible example for this problem. Because than I could reproduce this. I don't know what happens with action mailer when you try this. I will close it now.

gianmarcotoso:
Very well. But by doing this you are effectively NOT supporting SSL. Take a look a the official examples in GitLab documentation: https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/doc/settings/smtp.md 
The QQ exmail example uses a configuration that is currently not possible to achieve with your repository due to how your container is configured and how variables from the `docker-compose.yml` are reflected in the settings file. SSL and TLS are indeed two different protocols, and TLS has nothing to do with this PR.

I still think you have completely misunderstood the intention of this PR

PR #912, opened 2016-10-01
Switching from ReplicationController to Deployment.

grugnog:
Deployments use declarative configuration and have a simpler workflow (for most use cases, at least) for pushing updates than ReplicationController. It would be good to consider moving the sample k8s configurations over to this format.

### DIFF
--- a/kubernetes/deploy.sh
+++ b/kubernetes/deploy.sh
@@ -12,11 +12,11 @@ do
   kubectl -f $svc create
 done
 
-# create the replication controllers
-for rc in *-rc.yml
+# create the deployments
+for deploy in *-deploy.yml
 do
-  echo -n "Creating $rc... "
-  kubectl -f $rc create
+  echo -n "Applying $deploy... "
+  kubectl create -f $deploy --record
 done
 
 # list pod,rc,svc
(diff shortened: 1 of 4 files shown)


kubernetes/deploy.sh
```diff
@@ -12,11 +12,11 @@ do
...
 
-# create the replication controllers
-for rc in *-rc.yml
+# create the deployments
+for deploy in *-deploy.yml
 do
-  echo -n "Creating $rc... "
-  kubectl -f $rc create
+  echo -n "Applying $deploy... "
+  kubectl create -f $deploy --record
```
gambol99: perhaps kubectl 'apply' is a little more convenient than 'create'

sameersbn:
Thanks @grugnog. You are right, we need to switch to using deployments. I intend to replace/rewrite all the kubernetes manifests as a helm chart. It would be much easier to manage deployments on kubernetes.

Dm3Ch:
So when deployments will be in master branch?

rutsky:
@sameersbn 
> I intend to replace/rewrite all the kubernetes manifests as a helm chart.

FYI there is a helm chart based on your docker-gitlab images by @lwolf: https://github.com/lwolf/gitlab-chart

PR #914, opened 2016-10-01
WIP GitLab 8.12.2

solidnerd:


lenovouser:
This tag should not be released. From [gitlab.com/gitlab-8.12.3-released](https://about.gitlab.com/2016/09/29/gitlab-8-12-3-released/):

> Version 8.12.3 contains some security fixes for GitLab, plus fixes for minor regressions. Version 8.11.8, 8.10.11, and 8.9.11 only contain the security fixes.

> If you're wondering what happened to 8.12.2, good eye! That version was accidentally packaged without including some fixes for the CE version.

solidnerd:
Yeah thanks for hint I didn't saw the blog post about it. I will close this and work on 8.12.3.

xueshanf:
Is there a plan to release 8.11.8 security release? We aren't ready to upgrade to 8.12 release yet. Thanks!

solidnerd:
@xueshanf At the moment there are no plans to support older security releases. The problem because we don't make security releases is that GitLab by it self is multi branched for every major version. This allows them to make fixes to older version. We are single branched and can go only straight forward. When we want to do this in single branch we need to rewrite the commit history and that is not a good idea. One solution for you would be to make a fork from us and do it by your own. I can help you if you want. For a general support for security releases we should discuss this an separate issue.

xueshanf:
@solidnerd I see. I will make upgrade next few days. On the other hand, it would be good to have some docs in general on how to build the image with specific versions.  Great work in this project, much appreciated.

PR #920, opened 2016-10-07
Store gitlab-workhorse & mail_room stdout/stderr log files in correct directory.

tbeadle:
Make the stdout/stderr log files captured by supervisord for gitlab-workhorse
and mail_room to be stored in the same directory as the other child
processes.

### DIFF
--- a/assets/build/install.sh
+++ b/assets/build/install.sh
@@ -269,8 +269,8 @@ command=/usr/local/bin/gitlab-workhorse
 user=git
 autostart=true
 autorestart=true
-stdout_logfile=${GITLAB_INSTALL_DIR}/log/%(program_name)s.log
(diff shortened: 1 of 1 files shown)


solidnerd:
Thanks for this I will look later into it and merged it when it's LGTM . Otherwise I would give you feedback to you PR.

sameersbn:
@tbeadle The upstream instructions/scripts capture these two logs in the `log` directory of the GitLab installation. This is probably because these two daemons have no logging mechanisms other than logging to the stdout/stderr.

We prefer to have the gitlab related logs on the log/ directory of the gitlab installation and hence we log these to files in `GITLAB_INSTALL_DIR/log`.

I am of the opinion that this should not be changed.

PR #921, opened 2016-10-07
Allow supervisord to capture unicorn's stdout/stderr.

tbeadle:
Supervisord is already configured to log stdout/stderr for unicorn, but
nothing was getting to it because it was being directed to a different
place due to this config. Allow supervisord to handle it so that it's in
the same directory as the other processes managed by supervisord.


solidnerd:
Thanks for this I will look later into it and merged it when it's LGTM .

sameersbn:
Thanks for the contribution @tbeadle, but don't you think this change is unnecessary? I prefer to have the unicorn configuration intact and would also help in the event we decide to dump `supervisord` at some point.

PR #924, opened 2016-10-12
Add nginx retain ip header

gaoyifan:
If docker container runs behind a reverse proxy (like CDN), unicorn cannot get real client IP, because nginx overrides HTTP header `X-Forwarded-For` and `X-Real-IP`. 

`NGINX_RETAIN_IP_HEADER` parameter let nginx retain `X-Forwarded-For` and `X-Real-IP`.


knight42:
ping @sameersbn @solidnerd

solidnerd:
I will test this feature within a week and will give you feedback after that. I hope it's okay for you.

knight42:
@solidnerd Any progress?

lenovouser:
Seems like a good feature to me 👍

solidnerd:
Should this be done for [nginx-ssl](https://github.com/sameersbn/docker-gitlab/blob/master/assets/runtime/config/nginx/gitlab-ssl) as well ? Is it a possible test case ?

lenovouser:
Yep, should be done for the SSL part too IMO

knight42:
@solidnerd 
`nginx_configure_gitlab_retain_ip_header()` is called in `nginx_configure_gitlab()`, which means it directly modify `/etc/nginx/sites-enabled/gitlab`, hence I think there is no need for this to be done for `nginx-ssl`.

phenomax:
Do we have any updates here?
I also agree that this PR adds an useful feature to this docker image.

dmednis:
When can we expect this to be merged? 
This would be a really useful feature.

gaoyifan:
@dmednis  You can try [this fork](https://github.com/ustclug/docker-gitlab). Maintained by @knight42

3kami3:
I tried setting [GitLabCE official document](https://docs.gitlab.com/omnibus/settings/nginx.html#configuring-gitlab-trusted_proxies-and-the-nginx-real_ip-module), so please also try this PR #1137.

gaoyifan:
Agree with @3kami3 😀.  I think the implementation in #1137 is safer. And I will close this one.

PR #937, opened 2016-10-23
Improve container registry docs and certificate handling

maikelvl:
- Improvements to the registry documentation
- Reuse GITLAB_REGISTRY_KEY_PATH
- Reuse GITLAB_REGISTRY_CERT_PATH


assets/runtime/env-defaults
```diff
@@ -130,7 +130,9 @@ SSL_DHPARAM_PATH=${SSL_DHPARAM_PATH:-$GITLAB_DATA_DIR/certs/dhparam.pem}
 SSL_VERIFY_CLIENT=${SSL_VERIFY_CLIENT:-off}
 SSL_CIPHERS=${SSL_CIPHERS:-'ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA256:ECDHE-RSA-AES256-SHA:ECDHE-RSA-AES128-SHA:ECDHE-RSA-DES-CBC3-SHA:AES256-GCM-SHA384:AES128-GCM-SHA256:AES256-SHA256:AES128-SHA256:AES256-SHA:AES128-SHA:DES-CBC3-SHA:!aNULL:!eNULL:!EXPORT:!DES:!MD5:!PSK:!RC4'}
 
+SSL_REGISTRY_KEY_PATH=${SSL_REGISTRY_KEY_PATH:-$GITLAB_REGISTRY_KEY_PATH}
```
solidnerd: What is purpose of line because I will be ovewritten by line 134

assets/runtime/env-defaults
```diff
@@ -130,7 +130,9 @@ SSL_DHPARAM_PATH=${SSL_DHPARAM_PATH:-$GITLAB_DATA_DIR/certs/dhparam.pem}
...
 SSL_REGISTRY_KEY_PATH=${SSL_REGISTRY_KEY_PATH:-$GITLAB_DATA_DIR/certs/registry.key}
+SSL_REGISTRY_CERT_PATH=${SSL_REGISTRY_CERT_PATH:-$GITLAB_REGISTRY_CERT_PATH}
```
solidnerd: Same behaviour like my comment before

assets/runtime/functions
```diff
@@ -964,10 +964,11 @@ map_uidgid() {
 }
 
 update_ca_certificates() {
```
solidnerd: This could be a separated commit. To make an easier revert possible.

docs/docker-compose-registry.yml
```diff
@@ -0,0 +1,96 @@
...
+    - postgresql:/var/lib/postgresql:Z
+    environment:
+    - DB_USER=gitlab
+    - DB_PASS=password
+    - DB_NAME=gitlabhq_production
+    - DB_EXTENSION=pg_trgm
+
+  gitlab:
+    restart: always
+    image: sameersbn/gitlab:8.12.7-dev
```
solidnerd: Use current image tag `sameersbn/gitlab:8.12.7`

maikelvl: Very sharp. ;) Was used for testing.

remohammadi:
If I understood correctly, this is not just "Improve container registry docs". This commit will change the way registry is exposed to the user. In the current docs, the registry is hidden, and requests are passed through GitLab. But in your changes, registry is exposed to the user, and is set to use GitLab as the authentication driver. Is it the expected result?
@maikelvl @solidnerd

assets/runtime/env-defaults
solidnerd: I didn't say anything  😏 .

maikelvl:
@remohammadi Thank you for your comment. In my understanding the registry proxies the authentication to GitLab, so the registry has to be exposed to the user. I found the docs were not aligned with this, so I fixed it. If so, could you explain why this is harmful?

solidnerd:
@maikelvl You are correct. What GitLab does it will be used as authentication provider for the registry. You don't need to use the all in one solution. But this docs needed an update. I had no time for this and @maikelvl  did the great work.  This will be tested before we merge it. I want to do some other stuff because the impact  will not so hard for current users what I see.

remohammadi:
Thank you @maikelvl and @solidnerd . I was just checking my understanding. The title was a bit misleading; I didn't expect a change in behavior from that title :wink: 

So, should we reconfigure the previously configured dockers to match this configuration after these changes are released? Or are them backward-compatible?

maikelvl:
It is backwards compatible. 😄

docs/docker-compose-registry.yml
```diff
@@ -0,0 +1,96 @@
...
+    ports:
+    - "5000:5000"
+    volumes:
+    - registry-data:/var/lib/registry
+    - ./certs:/certs
+    external_links:
+    - "gitlab:gitlab.example.com"
+    environment:
+    - REGISTRY_LOG_LEVEL=info
+    - REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=/registry
```
solidnerd: And you mount your volumes to `/var/lib/registry`. I guess it should be preferred to have your images on your host :)

maikelvl: Yes, that's correct. Thanks. I fixed it now.

The registry data is persistent by storing it in a docker volume.

solidnerd:
Sorry I will use this now. Thanks for your work and patients ;) .

maikelvl:
Thanks. 😄

esben:
I know you said this change was backwards compatible, but my setup broke with the upgrade from 8.17.3 to 8.17.4.

I see

```
2017/03/22 07:15:41 [emerg] 3283#3283: SSL_CTX_use_PrivateKey_file("/home/git/data/certs/registry-auth.key") failed (SSL: error:0B080074:x509 certificate routines:X509_check_private_key:key values mismatch)
```
in /var/log/nginx/error.log

and I suspect it might be related to this change, but haven't figured it out yet...

solidnerd:
It's interessting. I have not theses problems.

esben:
In my case, I did not specify neither of SSL_REGISTRY_KEY_PATH, SSL_REGISTRY_CERT_PATH, and GITLAB_REGISTRY_CERT_PATH.

I guess my problem is caused by SSL_REGISTRY_KEY_PATH being set different (ie. to GITLAB_REGISTRY_KEY_PATH) than before this change.

And in that light, it is not completely fair to call this backwards compatible.  I have fixed it now, by specifying all of the 3 above mentioned variables, and are up and running with 8.17.4.

maikelvl:
Hi @esben I'm sorry to hear it's not working for you and indeed not correct of me to call it backwards compatible. In most cases it would be backwards compatible. 😅

Good to hear you fixed it.

gdoteof:
It looks to me like gitlab_registry_key_path and ssl_registry_key_path are actually the same thing?

Very confusing to me in the docs listed here:  https://github.com/sameersbn/docker-gitlab/blob/master/docs/container_registry.md

```
docker run --name gitlab -d [PREVIOUS_OPTIONS] \
-v /srv/gitlab/certs:/certs \
--env 'SSL_REGISTRY_CERT_PATH=/certs/registry.crt' \
--env 'SSL_REGISTRY_KEY_PATH=/certs/registry.key' \
--env 'GITLAB_REGISTRY_ENABLED=true' \
--env 'GITLAB_REGISTRY_HOST=registry.gitlab.example.com' \
--env 'GITLAB_REGISTRY_API_URL=http://registry:5000/' \
--env 'GITLAB_REGISTRY_CERT_PATH=/certs/registry-auth.crt' \
--env 'GITLAB_REGISTRY_KEY_PATH=/certs/registry-auth.key' \
--link registry:registry
sameersbn/gitlab:10.0.4
```

the above sets both, with different values.  the /certs/registry-auth.* versions are both shown how to be generated in the documentation, but there is no mention of /certs/registry.*

PR #941, opened 2016-10-27
Add Introduced In column to Available Configuration Parameters

faultylee:
Thanks for the great works in maintaining this!!!

I usually only upgrade when the next next minor release is out, i.e. I'm upgrading to 8.12.7 (from 8.11.7) when 8.13.0 is out, to avoid having to down Gitlab too often and also having to deal with minor bugs. I constant find myself having to monitor the issues to know if there's any newly added environment variable. Hence I thought of asking you guys to add a column but I end up doing it myself, though I've only done it for the recently added one for 8.13.0. I think it's a good start. Feel free to amend or correct any mistake I might have :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -790,204 +790,204 @@ These options should contain something like:
 
 Below is the complete list of available options that can be used to customize your gitlab installation.
 
-| Parameter | Description |
-|-----------|-------------|
-| `DEBUG` | Set this to `true` to enable entrypoint debugging. |
-| `GITLAB_HOST` | The hostname of the GitLab server. Defaults to `localhost` |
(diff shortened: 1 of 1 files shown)


faultylee:
Btw, it did mess up the Description column a little (too wide), not sure how I can fix that though

solidnerd:
At the moment I don't see  why we should add further column. Newly added variables are listed in [Changelog.md](https://github.com/sameersbn/docker-gitlab/blob/master/Changelog.md) .

faultylee:
Sorry, my mistaken, didn't notice that. I saw the same thing from github's email notification, but didn't realized it's from there. In that case maybe a note/link to `changelog.md` above the table might be helpful.

Thanks again, I'll close this.

Issue #945, opened 2016-11-03
HTTPS problem on Synology NAS DS916+


heximcz:
Hello,

I've problem with https. Http working without problem, but when I run image with:
```
...
    --publish 30022:22 --publish 30080:80 --publish 30443:443 \
    --env 'GITLAB_PORT=30443' \
    --env 'GITLAB_HTTPS=true' \
...
```
```
Certificate is in :
root@gitlab:/home/git/gitlab# ls -la /home/git/data/certs/
-rw-r--r-- 1 git git  424 Nov  2 17:43 dhparam.pem
-rw-r--r-- 1 git git 3461 Nov  2 18:25 gitlab.crt
-r-------- 1 git git 1679 Nov  2 18:25 gitlab.key
```
```
Terminal output (docker log):
2016-11-03 00:27:49,883 INFO spawned: 'nginx' with pid 3152                                                                                                   
2016-11-03 00:27:50,911 INFO success: nginx entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)                                     
2016-11-03 00:27:52,414 INFO exited: nginx (exit status 1; not expected)                                                                                      
2016-11-03 00:27:53,418 INFO spawned: 'nginx' with pid 3153                                                                                                   
2016-11-03 00:27:54,448 INFO success: nginx entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)                                     
2016-11-03 00:27:55,950 INFO exited: nginx (exit status 1; not expected)                                                                                      
2016-11-03 00:27:56,953 INFO spawned: 'nginx' with pid 3154
```
```
tailf /var/log/nginx/error.log :

2016/11/03 00:30:53 [emerg] 3264#3264: bind() to [::]:80 failed (98: Address already in use)                                                                  
2016/11/03 00:30:53 [emerg] 3264#3264: bind() to [::]:443 failed (98: Address already in use)                                                                 
2016/11/03 00:30:53 [emerg] 3264#3264: bind() to [::]:80 failed (98: Address already in use)                                                                  
2016/11/03 00:30:53 [emerg] 3264#3264: bind() to [::]:443 failed (98: Address already in use)                                                                 
2016/11/03 00:30:53 [emerg] 3264#3264: bind() to [::]:80 failed (98: Address already in use)                                                                  
2016/11/03 00:30:53 [emerg] 3264#3264: bind() to [::]:443 failed (98: Address already in use)                                                                 
[...]

PR #950, opened 2016-11-03
fix nginx ipv6only off

heximcz:
Fix for #945


solidnerd:
You were faster 😄 . I was atm to start working on the fix. Thanks for your work. I will be available on `latest`and later in `8.13.3` but this needs to be tested first.

heximcz:
On Synology NAS it working fine. :-)

Issue #951, opened 2016-11-03
Fail HTTP redirect to HTTPS


heximcz:
HTTP Port = 30080
HTTPS Port = 30443

wget test redirect:
```
wget http://gitlab.<domain>:30080 -O ./test.txt
...
Connecting to gitlab.<domain> (gitlab.<domain>)| x.x.x.x|:30080... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: https://gitlab.<domain>:30080:30443 [following]
https://gitlab.<domain>:30080:30443: Bad port number.
...
```

solidnerd:
It's merged . I wil be available in `latest`and later in `8.13.3`

PR #952, opened 2016-11-03
fix http to https redirect

heximcz:
Fail HTTP redirect to HTTPS #951


GottZ:
http://stackoverflow.com/a/15414811/1519836

- $http_host equals always the HTTP_HOST request header.
- $host equals $http_host, lowercase and without the port number (if present), except when HTTP_HOST is absent or is an empty value. In that case, $host equals the value of the server_name directive of the server which processed the request.

so i guess you want to avoid ports right?

heximcz:
Yes right. Because if I use non standard port as 30080 for http, redirect will be:
```
https://<domain>:30080:30443 (bad port number)
```
More in issue #945

solidnerd:
Thanks 👍  . LGTM ! I will merge this .

Issue #944, opened 2016-11-02
Security update: GitLab 8.13.3 scheduled for release


evenh:
Hey!

Just wanted to raise awareness of GitLab's pending release of 8.13.3 at Wednesday November 2, 2016 at 4:59pm PDT (23:59 GMT), due to CVE-2016-9086.

For more details, see the [GitLab blog](https://about.gitlab.com/2016/10/31/gitlab-major-security-update-for-cve-2016-9086/).

solidnerd:
Hey,
thanks for the info. I will prepare a release for it when the tag is available.

PR #948, opened 2016-11-03
Update to v8.13.3

reynold-lariza:
Attached issue #944

PR #953, opened 2016-11-04
Upgrade to GitLab 8.13.3

evenh:
Here is a single commit upgrade :-)

Ref. #944


solidnerd:
Sorry I will use #948 . It's okay for that are are two commits, because they do different things.

PR #984, opened 2016-11-26
Fix MySQL first run (a foreign key constraint fails)

xoob:
Fixes an issue where starting a new MySQL based container would fail
due to foreign key constraints.

To avoid this problem, the order of steps is changed, so `gitlab:setup` runs before `db:migrate`.

```
gitlab_1  | Migrating database...
gitlab_1  | Setting up GitLab for firstrun. Please be patient, this could take a while...
gitlab_1  | gitlabhq_production already exists
gitlab_1  | rake aborted!
gitlab_1  | ActiveRecord::StatementInvalid: Mysql2::Error: Cannot delete or update a parent row: a foreign key constraint fails: DROP TABLE `boards` CASCADE
...
gitlab_1  | /home/git/gitlab/lib/tasks/gitlab/setup.rake:17:in `setup_db'
gitlab_1  | /home/git/gitlab/lib/tasks/gitlab/setup.rake:4:in `block (2 levels) in <top (required)>'
...
gitlab_1  | Tasks: TOP => db:schema:load
gitlab_1  | (See full trace by running task with --trace)
gitlab_gitlab_1 exited with code 1
```

With this fix, it works correctly:

```
gitlab_1  | Setting up GitLab for firstrun. Please be patient, this could take a while...
gitlab_1  | gitlabhq_production already exists
gitlab_1  | Migrating database...
gitlab_1  | Clearing cache...
gitlab_1  | 2016-11-26 12:16:27,366 CRIT Supervisor running as root (no user in config file)
...
```

### DIFF
--- a/assets/runtime/functions
+++ b/assets/runtime/functions
@@ -1294,6 +1294,13 @@ migrate_database() {
       ;;
   esac
 
+  if [[ -z ${COUNT} || ${COUNT} -eq 0 ]]; then
+    echo "Setting up GitLab for firstrun. Please be patient, this could take a while..."
+    exec_as_git force=yes bundle exec rake gitlab:setup \
+      ${GITLAB_ROOT_PASSWORD:+GITLAB_ROOT_PASSWORD=$GITLAB_ROOT_PASSWORD} \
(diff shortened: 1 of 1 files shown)


sameersbn:
thanks for the PR @xoob 
I am testing this PR. Is there an issue reported that is solved by this PR?

PR #989, opened 2016-11-29
Upgrading to Workhorse 1.0.1 - Fixes missing version tag

phenomax:
As explained by GitLab [here](https://gitlab.com/gitlab-org/gitlab-ce/commit/5981a0f365b9a312b87bfcebfb0d9365cb2f9c9a) the upgrade to v1.0.1 fixes the missing version tag for workhorse @reynold-lariza has already discovered when testing GitLab v8.14.0 in [this](https://github.com/sameersbn/docker-gitlab/pull/979) pull request

PR #994, opened 2016-12-02
Upgrade GitLab to 8.14.2

solidnerd:
Hey,

i prepared the upgrade for `8.14.2` this includes the update for workhorse. So there will be the mission version fixed. 

Closes #989


solidnerd:
Waiting for a sync of https://github.com/gitlabhq/gitlabhq .

Ping @smcgivern could you let this happen ?

smcgivern:
That's done now!

Just out of curiosity, could you use the GitLab.com repo instead? https://gitlab.com/gitlab-org/gitlab-ce/tags

That's guaranteed to be up-to-date 😃

phenomax:
You have to upgrade workhorse to v1.**1**.0 instead of v1.**0.1**. 🔢 
[Workhorse changelog](https://gitlab.com/gitlab-org/gitlab-workhorse/blob/master/CHANGELOG)

solidnerd:
@smcgivern Last time we switched to gitlab.com we got some problems with shallow cloning. Here is a pr for that #842 .

I have a second question why is the github repo not automically mirrored from gitlab. This could be done by a CI Job 😄  .

But thanks anyway for the work 😄  .

smcgivern:
@solidnerd fair enough! Just bear https://github.com/gitlabhq/gitlabhq#canonical-source in mind 😉 

I don't know why we don't mirror, but we do push to GitHub as [part of our release process](https://gitlab.com/gitlab-org/release-tools/blob/master/doc/push-to-multiple-remotes.md) - it just got missed for some reason on this patch. Sorry about that!

Issue #995, opened 2016-12-02
All master branches are protected with upgrade to 8.14.1


adamlc:
Not sure what's happened but after upgrading from 8.14.0 to 8.14.1 all the master branches in all repos have had their status changed to protected, maybe more bad migration or something? Anyone else had this issue? All other branches seem to be OK.

I know its not the end of the world because changing back to unprotected works fine, but thought I'd mention it 👍

r2evans:
That is the default behavior for GitLab (src: [protected_branches.html](https://docs.gitlab.com/ce/user/project/protected_branches.html#configuring-protected-branches)). I don't know when it became default, since a [commit on July 22](https://gitlab.com/gitlab-org/gitlab-ce/commit/1a5b7e7d70cfe4802e858f129779e388307fbbab#ce29b2e1034f54526853bdd8fb7cdc4e3f2757d8) implies the verbiage (on line 13 of the old version) was already there at that time.

I'm still on `sameersbn/gitlab:8.14.0` and my `master` branches are all protected (and I've done nothing to set or change them).

I'm not saying this is not an issue. If you had previously *un*protected all of your `master` branches, or the branches were created before this became the default mode in GitLab, then who knows.

If that's the case, then it's likely something for [GitLab issues](https://gitlab.com/gitlab-org/gitlab-ce/issues/), not here, since I believe this docker image is not doing anything *that specific* to repos.

gilyes:
I upgraded as well from 8.14.0 to 8.14.1, and now I can't push to my protected branches that I could push to before. Getting error saying "`remote: GitLab: You are not allowed to force push code to a protected branch on this project`" even though I am not force pushing. But yeah, most likely this is an issue for GitLab.

r2evans:
Users belonging to the "Masters" list can push to the `master` branch. I think only the repo owner (since it is a per-repo setting) can change this, but perhaps other "Masters" can do it as well.

To find the list, within the repo pull down the "repo gear" (top-right corner, typically under the user avatar) and select "Members". Each member that has been granted access will be specifically listed as a "Master" (among "Guest", "Reporter", "Developer", and "Master"). If instead of "Members" you go to "Protected Branches", the level required to push can be set. "Merge" ops can be "Masters" or "Masters+Developers", while "Push" ops can be restricted to "Masters", "Master+Developers", or "No one".

[...]

PR #1013, opened 2016-12-12
Downgrade to git 2.10.2.

lyda:
(tagging #995 )

Due to issue
[995](https://github.com/sameersbn/docker-gitlab/issues/995)
which in turn is due to issue
[25301](https://gitlab.com/gitlab-org/gitlab-ce/issues/25301)
git version 2.11.0 (a recent ubuntu 14.04 upgrade) is not working
correctly.  This removes the git-core ubuntu package install
and replaces it with a git install from source. It does so
in an easily removable block.

Note that the image will be a bit larger than normal.  The
built git needs some libraries installed and I didn't spend
the time figuring out which.


adamlc:
Any updates on this? Would love to get this fixed, even if its only temporary :)

solidnerd:
@adamlc Sorry for the delay. I updated the gitlab version to 8.15+ and it should work now. Could you upgrade your GitLab version and say if it works ?

lyda:
I upgraded and it works fine. This is no longer required.

PR #1015, opened 2016-12-12
Describe virgin restores.

lyda:
Restores on a system with no existing database need an additional
step. In addition the gitlab backup does not backup the `ssh/`
directory which is where the host keys for the ssh server are
kept.


solidnerd:
LGTM !

PR #1049, opened 2017-01-10
Mention official Docker image for GitLab in README

adamniedzielski:
Hey! Thanks for creating this project, it seems that a lot of people are using it! GitLab now has its official Docker images and official instructions for using them. I think that it's fair to mention it in the README, so people are aware that they use an unofficial installation method. What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@
 - [Issues](#issues)
 - [Announcements](https://github.com/sameersbn/docker-gitlab/issues/39)
 - [Prerequisites](#prerequisites)
+- [Standard Docker image](#standard-docker-image)
 - [Installation](#installation)
 - [Quick Start](#quick-start)
 - [Configuration](#configuration)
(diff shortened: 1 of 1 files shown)


reynold-lariza:
Interesting! I didn't know there was an official docker image of Gitlab! :O

dashersw:
Unofficial sounds "fake". I reviewed the "original" Docker images too; but still, this method is better.

adamniedzielski:
@dashersw

> Unofficial sounds "fake".

Feel free to suggest a better wording!

> I reviewed the "original" Docker images too; but still, this method is better.

Yeah, everybody is free to decide on their own 😄. Being an "official image" means that people contributing to GitLab are less likely to break it (because they consider it when making a change). That's why I think that it makes sense to let people know about it.

dashersw:
I don't think this repo should "let people know about it", it's GitLab's responsibility to do that. Make sure to put it in bold in GitLab setup documentation, and people won't have to search Google for docker gitlab.

This repo is updated in a controlled way; so it doesn't really break. I'm using it since 2014 and I don't remember it ever breaking. The only problem is, the releases follow GitLab's releases after a few days, because it takes time to update and verify this repository.

adamniedzielski:
@dashersw Thank you very much for suggesting a better wording! I've updated the PR.

I think that the most obvious way to learn about installation methods for GitLab is https://about.gitlab.com/installation/. And this page mentions "Docker installation instructions for a container based on the Omnibus packages". We cannot really put it in bold there, because this would suggest that it's somehow more important than other methods which is not the case. Especially that the recommended way is still Omnibus package installation.

Anyway, I'm leaving it up to the maintainers of this repo whether to accept this PR or close it. Obviously it's their decision what they want to have in the README 😺.

dashersw:
No problem :) I would merge it this way :)

sameersbn:
@adamniedzielski @dashersw Sorry for my lateness in responding to your PR. I think this is a good idea. I've created #1060 with the required changes. If it's acceptable to you, I get it merged. Best Regards.

PR #1084, opened 2017-02-09
Changing docker compose example

lchigami:
Fixing docker compose example. The original values doesn't work.
The port 5500 isn't a exposed port in the Gitlab container.
The link between Gitlab and Registry containers is necessary for GITLAB_REGISTRY_API_URL to be able to resolve the name "registry" in the URL.

### DIFF
--- a/docs/container_registry.md
+++ b/docs/container_registry.md
@@ -100,9 +100,9 @@ services:
     depends_on:
     - redis
     - postgresql
+    - registry
     ports:
     - "10080:80"
-    - "5500:5500"
     - "10022:22"
     volumes:
     - ./gitlab:/home/git/data:Z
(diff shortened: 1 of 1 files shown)


brunosavyofs:
I've been struggling with these confs for days because of this wrong port configuration in docker-compose.yml.

docs/container_registry.md
```diff
@@ -136,6 +136,8 @@ services:
   registry:
     restart: always
     image: registry:2.4.1
+    ports:
+    - "5000:5000"
```
dfeyer: This is not required. And gitlab export the port and proxy the access to the registry.

solidnerd:
This isn't needed any more in favor of #937 .

PR #1087, opened 2017-02-15
Updates to Gitlab 8.16.5

iamchanghyunpark:
Changelog at [Gitlab blog](https://about.gitlab.com/2017/02/15/gitlab-8-dot-16-dot-5-security-release/)

Build Successful: [Docker hub](https://hub.docker.com/r/heartinpiece/gitlab/tags/)
Successful Launch:
image

Signed-off-by: Chang Hyun Park <heartinpiece@gmail.com>

PR #1089, opened 2017-02-16
Upgrade to 8.16.5

kristiandrucker:


solidnerd:
Thanks for PR but it seems this is a duplicate of #1087.

PR #1091, opened 2017-02-18
Backups fix logical check of booleans

paolomainardi:


paolomainardi:
Sorry but the current check fail if AWS_BACKUPS is false, it doesn't evaluate the second argument, sorry for my poor bash skills

paolomainardi:
cc @solidnerd

solidnerd:
LGTM ! I didn't see that in the first PR. Thanks for contrib.

PR #1111, opened 2017-03-02
Allow setting sshd MaxStartups with env variable (fixes #969)

### LINKED ISSUE #969: Error ssh_exchange_identification because of too many connections
Is there a way I can edit the sshd configuration to include

```
MaxStartups 60
MaxSessions 60
```

So ssh works the way I need it to?

jprjr:


jprjr:
I probably should have added some background on why I did this back when I opened this!

We host private NPM modules in our Gitlab and deploy via git-over-ssh. NPM tends to open a lot of simultaneous connections to gitlab, being able to set the MaxStartups item is really helpful for that use-case.

derphilipp:
Almost the same here, lot's of mirrored C/C++ Packages that are used as submodules;
As those are quite large, settings "--jobs 32" for git clone/submodule processing, helps a lot in terms of speed. Unfortunately gitlab does not accept enough connections and therefore fails;

xsteadfastx:
This is pretty useful!

derphilipp:
Also when running several parallel builds, my builds tend to break due to the fact that the not enough ssh connections can be opened up at once. This would solve it

xelatirdan:
@solidnerd Could you merge this helpful fix?

zwopir:
@solidnerd is it possible to get this one merged (or a similar, non-conflicting one I can create if the original author of this PR is not available). As stated in https://github.com/sameersbn/docker-gitlab/issues/969#issuecomment-526900663 the official docs recommend tweaking the `MaxStartups` parameters if users run into the the `ssh_exchange_identification connection closed by remote host` error.

jprjr:
Hi there, I fixed the merge conflicts.

Issue #1112, opened 2017-03-04
Upgradability - docker


brubakerjeff:
Hi,

Would it be better to add more volumes to https://github.com/sameersbn/docker-gitlab/blob/master/docker-compose.yml as shown in https://docs.gitlab.com/omnibus/docker/#upgrade-gitlab-to-newer-version to enable upgradability?
 
Thanks,
Jeff

gitlab:
    restart: always
    image: sameersbn/gitlab:8.16.6
    depends_on:
    - redis
    - postgresql
    ports:
    - "10080:80"
    - "10022:22"
    volumes:
instead of    
- /srv/docker/gitlab/gitlab:/home/git/data:Z

use
  - /srv/gitlab/config:/etc/gitlab \
   - /srv/gitlab/logs:/var/log/gitlab \
   - /srv/gitlab/data:/var/opt/gitlab \**

solidnerd:
The difference between this image and the official one is that we don't use omnibus for configuration and install from source. We have the same upgradeability like the official one. You need only to change your image tag to the next version.

ragaar:
@solidnerd, I have found it is also mandatory to create a storage volume to `/home/git/data`.

I've been on `8.17.4` for awhile, and only recently came back to this. I was able to upgrade to `9.0.0` but it required `/home/git/data` to be persistent.

nixel2007:
@ragaar all gitlab data stored at /home/git/data. So, yes, if you want to recreate/upgrade your container, you need to save this directory

PR #1113, opened 2017-03-04
As suggested

phpdave:
https://github.com/sameersbn/docker-gitlab/issues/1112


solidnerd:
This configuration has no impact for the image. We do this in a different way like the official image because we use only sources and no omnibus. So you volumes won't be used.

PR #1126, opened 2017-03-15
add GITLAB_PORT_EXT, to cover cases where GITLAB_PORT is actually not the port the service is being accessed from

neta79:
This is necessary in those cases where GITLAB_PORT differs from the port used by a load balancer or proxy to publish the service to the outside world. Without this mapping GitLab generates wrong backlinks to itself.


solidnerd:
This is not needed I you want to talk directly to gitlab-workhorse expose `8181` and let you load balancer handle it.

monkeydri:
@solidnerd could you elaborate please ?

I run this docker-gitlab behind a nginx reverse-proxy (linuxserver/letsencrypt docker) so only 443 is exposed on host.

The thing is internally the port used by the gitlab instance for http is sometimes appended to the images URLs (therefore the images are not displayed because this port is not opened on host).

neta79:
@solidnerd 
@monkeydri described precisely our same scenario. 
Gitlab generates urls with the only port info it is aware of. 
In our case, this patch keeps being essential to avoid occasional bad urls. 

Please reconsider merging.

solidnerd:
@monkeydri @neta79  You don't need this. What you can do is to that the `GITLAB_PORT` and it will solve your problems. I also run GitLab behind a proxy and it's works for me.

PR #1137, opened 2017-03-20
Add $NGINX_REAL_IP_RECURSIVE and $NGINX_REAL_IP_TRUSTED_ADDRESSES

3kami3:
Configuring GitLab trusted_proxies and the NGINX real_ip module.
If your docker container runs behind a reverse proxy, you may not want the IP address of the proxy to show up as the client address.
ref.
https://docs.gitlab.com/omnibus/settings/nginx.html#configuring-gitlab-trusted_proxies-and-the-nginx-real_ip-module
and
https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/9623


xueshanf:
I am confused. I have been using GITLAB_TRUSTED_PROXIES. Is it going away?

3kami3:
I think that you can see in the description of the official GitLab document.

Using a LB and GitLab with bundled Nginx:
Set NGINX_REAL_IP_RECURSIVE and NGINX_REAL_IP_TRUSTED_ADDRESSES.
https://docs.gitlab.com/omnibus/settings/nginx.html#configuring-gitlab-trusted_proxies-and-the-nginx-real_ip-module

Using a non-bundled web-server:
Set GITLAB_TRUSTED_PROXIES.
https://docs.gitlab.com/omnibus/settings/nginx.html#using-a-non-bundled-web-server

avluis:
😻 I need this~
The issue is that when using a load balancer, there is still an internal IP that is logged:
`LB -> GitLab NGINX -> GitLab`

reynold-lariza:
Ah I did a similar thing with #1223, I had to make this work because of Rack Attack. 

`AWS ELB > nginx-proxy > docker-gitlab-with-nginx`

However in my case I've had it set default to `0.0.0.0/0` since its just the nginx inside gitlab container.

solidnerd:
@3kami3 I will add this for the next feature (9.4.0). Thanks for your work 👍  .

3kami3:
Thank you for the merge.

PR #1139, opened 2017-03-20
Use yarn to install JS dependencies

phpb-com:
Since yarn is part of the image now, we should use it to install JS dependencies.


solidnerd:
👍

solidnerd:
LGTM ! I'm a little bit confused because in the upstream are two ways to do this the upgrade guide says use yarn and the installation guide says use npm. But I prefer yarn because we add this to the image. Thanks for your work.

PR #1146, opened 2017-03-22
Fix faulty nginx config for pages

phpb-com:
This will fix issues starting container when pages enabled with https support. Also will remove issues in some configurations where ipv6only will cause listen directive to be considered as duplicate by nginx.


phpb-com:
🥇 And I get the first run of Travis CI 😃

pmoralesp:
I can confirm this solves the problem :+1:

solidnerd:
LGTM !

Issue #1148, opened 2017-03-23
Upgrade to 9.0


dereknex:
Gitlab 9.0 has been released. https://about.gitlab.com/2017/03/22/gitlab-9-0-released/

andrewstuart:
I've gotten my configuration at least working in my fork, so I'll at least submit a PR for those minor updates. Mostly I just followed the docs [here](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/doc/update/8.17-to-9.0.md) or updated the versions in the Dockerfile environment variables to the latest tags on Gitlab.

andrewstuart:
Although it seems I missed the pre-existing PR #1147, which is getting much more activity, so I'll probably just contribute to that one and close mine. :-)

dereknex:
I missed the PR #1147. 
Close this issue now.

PR #1147, opened 2017-03-22
Update Gitlab to 9.0.0

phpb-com:
This PR should cover all of the necessary changes for the update to Gitlab 9.0.0

PR #1149, opened 2017-03-23
Initial update to gitlab components for 9.0

andrewstuart:
I believe this should work for most existing configurations. At the very least, I was able to get it to start up, run my Kubernetes-based CI jobs, and deploy pages.

Addresses #1148


andrewstuart:
Closed in favor of #1147

PR #1166, opened 2017-04-06
Upgrade to GitLab CE 9.0.4

frostieDE:


frostieDE:
It seems that the commit itself did not cause the build error. Can someone try to retry the build, please? :-)

solidnerd:
Thanks for the PR . It seems like a hick up. So I restarted the test. But what I missing is the PR for an `9.0.3` version.

frostieDE:
I guess I can understand that `9.0.3` is needed to have a complete list of versions, but shouldn't we skip this version as it was superseded at the same day by `9.0.4` (which seems to be a security update)?

frostieDE:
I just integrated the `9.0.3` branch so you should not have any conflicts :wink:

grossjo:
Can we get this merged in?

PR #1174, opened 2017-04-12
fix version inconsistency in DB restore task doc

codereviewvideos:
Unsure if this is a mistake or not, but if not, likely needs some documentation as to why.

I'm new to your project, and docker in general, so apologies if I have missed something here. Thank you very much for sharing this project btw. I really appreciate it.


solidnerd:
Yeah 👍  .

LGTM !

PR #1196, opened 2017-04-27
Add tutorial for Synology setup

cpoetter:
Synology’s basic GitLab package is based on this Docker. But GitLab Runner or GitLab Container Registry are not available through the Package Center. Therefore I created a tutorial. Because this repository was my first location to look for an explanation how to set it up, I would like to link it in the ReadMe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@
     - [Shell Access](#shell-access)
 - [Features](#features)
  - [Container Registry](docs/container_registry.md)
+- [Synology](#synology)
 - [References](#references)
 
 # Introduction
@@ -1242,6 +1243,10 @@ For debugging and maintenance purposes you may want access the containers shell.
 ```bash
 docker exec -it gitlab bash
 ```
+
+# Synology
+
+To setup GitLab with Synology DSM follow this [tutorial](https://github.com/cpoetter/Synology-GitLab-Setup).
 
 # References
 


jostyee:
I'd rather commit the tutorial here instead of a link.

cpoetter:
@jostyee Thanks for your feedback! I also thought about that. But actually most of the information summarised in my tutorial is already part of this repo and [docker-gitlab-ci-multi-runner](https://github.com/sameersbn/docker-gitlab-ci-multi-runner) repo. It just needed to be slightly adjusted. For example the port for the registry can't be 5000, because that is the port used by the Synology DSM itself. Or the runner can't run in Docker executor mode, because a DSM user has no access to `/var/run/docker.sock`. Therefore the *gitlab_ci_multi_runner* user in the runner container should at least have the possibility to execute sudo commands if needed.

Therefore I did not want to duplicate the awesome documentation here, just pick the important part for a Synology user and write it somewhere up. Looking in to the [issues](https://github.com/sameersbn/docker-gitlab/issues?utf8=✓&q=Synology) convinced me to link it directly here in this repo.

GJRTimmer:
@cpoetter I'm running also on Synology, here is some additional information:

I've created a new gitlab-runner which is Synology compliant, the latest version of the runner of gitlab does not match the docker-engine version which is running on Synology. In order to run match version you need a runner which has docker-engine 1.11.2 as backend because of the internal docker-api.
The image does use the latest gitlab-runner.

See: https://hub.docker.com/r/datacore/gitlab-runner/
Repo: https://github.com/GJRTimmer/docker-gitlab-runner
Branch: Synology
Or: tag 1.11.2
Image is based on Alpine.

Additonally, maybe it is an idea to also write a section how to use the the Synology Certificate center in combination with gitlab to run with TLS/SSL enabled to use Let's encrypt certificates.

I've also created a image for building GoLang projects maybe a we could add a section for image references to point to build containers for building projects with Gitlab CI, what do you think ?

Golang image for CI: https://hub.docker.com/r/datacore/alpine-golang

cpoetter:
@GJRTimmer Awesome idea to use the Synology Certificate Center with Let's encrypt to provide certificates for GitLab. The certificates are currently the only section in my tutorial where users need to actually run terminal commands instead of the WYSIWYG DSM setup. It took me quite some time to configure the setup as is, but I will definitely look into that!

Regarding your runner Docker image. I was actually hoping to stay with @sameersbn runner version to keep it consistent. And as far as I can see a [pull request](https://github.com/sameersbn/docker-gitlab-ci-multi-runner/pull/33) which is updating to the newest runner version is going to merge soon. I tested the merged repo and it works fine on my DS916+ with the Synology GitLab version.

@jostyee These discussions are actually the reason I wanted to relocate the tutorial into a new repo. How to setup the Synology Certificate Center has nothing to do with this repo.

cpoetter:
@GJRTimmer I looked a bit into the certificate topic just now. The Certificate Center is storing the certificates in `/usr/syno/etc/certificate/_archive/`. The problem is, that this directory is not visible in the File Station Web UI. So we can't select it as a directory to be mounted into a Docker. Also unfortunately File Station does not support symbolic links. So linking the files into a directory that we can mount does not help.

This is the same issue as with the `/var/run/docker.sock` file for the Docker executor mode of the GitLab runner. I don't have a solution for this problem at the moment (besides launching the Docker via the terminal, which is not the intended solution for my tutorial. I want to write a tutorial that works hopefully just with the Web UI. I made already a pull request in the runner repo to get closer to my goal.).

I submitted already a feature request to support symbolic links in File Station. Maybe it would be a great idea if you do the same.

GJRTimmer:
Hi chris,

I'm running it with let's encrypt. Console acces is something I found
mandatory for this because of the use of docker-compose. I write a bash
script which makes everything quite easy. Including finding the correct
certificates from synology even after they are auto refreshed by the dsm.
Will upload the script soon.

Op vr 28 apr. 2017 17:52 schreef Chris <notifications@github.com>:

cpoetter:
@GJRTimmer I actually hope we can eliminate the need of console access soon.

- To install and configure GitLab itself we obviously don't need console access.
- To install and configure GitLab Runner we also do not need console access. Only if we want to enable *sudo* commands in the `gitlab-ci.yml` file. And to fix that I submitted a [pull request](https://github.com/sameersbn/docker-gitlab-ci-multi-runner/pull/37).
- To install and configure GitLab Container Registry we only need the console to create self signed certificates. This could also be done automatically while the registry Docker is booting up for the first time with the help of an environment variable flag.

This would of course only be the first step. The second step would be to find a solution to reuse the Synology Certificate Center certificates.

cpoetter:
@GJRTimmer I found a way to use the *Let's Encrypt* certificates without console access. We can use Synology's Reverse Proxy service as a load balancer with SSL termination. I added a new section regarding that in my [tutorial](https://github.com/cpoetter/Synology-GitLab-Setup#https).

I will look into the certificates used for the communication between the registry container and GitLab now. I think the easiest solution is to create a new pull request for this repository and add an environment variable flag to generate these certificates while booting the GitLab container the first time.

solidnerd:
Hey, 

why I would appreciate it as well to have this doc in our docs folder. So we have all on a single place. This seems for me a lot of work that you did to get it running on your synology. So WDYT ?

PR #1210, opened 2017-05-12
Switch to Ubuntu Xenial base image

israelshirk:
The diff should read pretty straightforwards; the only questionable part is regarding tzinfo-data:

```diff
+exec_as_git echo | tee -a Gemfile
+exec_as_git echo "gem 'tzinfo-data'" | tee -a Gemfile
+bundle install -j$(nproc)
```

I'm not a Rails ninja, but it did get things working.  Running the bundle install privileged required sudo, and running it as root is frowned upon.  Suggestions there would be very welcome.  Thanks!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,11 @@
-FROM sameersbn/ubuntu:14.04.20170503
+FROM ubuntu:xenial
 MAINTAINER sameer@damagehead.com
+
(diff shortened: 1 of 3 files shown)


solidnerd:
We won't change the base image at the moment sorry. For it would be alpine and not a new ubuntu version.

PR #1211, opened 2017-05-13
[WIP] Kubernetes updates

israelshirk:
Updates RCs to deployments and does some SOC on volumes, configs, and secrets to make production deployments easier.


solidnerd:
Hey, 
this looks quite good. Would you like to write an readme how to use it directly on kubernetes with an little `README.md`

solidnerd:
It seems that it's in WIP so I will add the tag to the title.

israelshirk:
Thanks!  I'm going to go through and add the remaining variables and fixes to get some level of parity/completeness to the underlying build this Thursday (USA/Mountain time); after that I think we'll be good!  I've been running it in production with SSL for about 5 days now and things have been going quite nicely.

solidnerd:
@israelshirk ping ! What's your current status of this pr ?

kubernetes/gitlab.yml
```diff
@@ -0,0 +1,253 @@
...
+        - name: http
+          containerPort: 80
+        - name: ssh
+          containerPort: 22
+        volumeMounts:
+        - mountPath: /home/git/data
+          name: data
+        livenessProbe:
+          httpGet:
+            path: /
```
yuanying: This livenessProve doesn't work because "/" responds "404".
So this path should be "/-/liveness" and to set "GITLAB_MONITORING_IP_WHITELIST" is needed.

kubernetes/gitlab.yml
```diff
@@ -0,0 +1,253 @@
...
+            port: 80
+          initialDelaySeconds: 900
+          timeoutSeconds: 5
+        readinessProbe:
+          httpGet:
+            path: /
```
yuanying: This also should be "/-/readiness".

kubernetes/gitlab-config-template.yaml
```diff
@@ -0,0 +1,35 @@
...
+  SMTP_PORT: "587"
+  SMTP_USER: "mailer@example.com"
+  SMTP_STARTTLS: "true"
+  SMTP_AUTHENTICATION: "login"
+  IMAP_ENABLED: "false"
+  IMAP_HOST: "imap.gmail.com"
+  IMAP_PORT: "993"
+  IMAP_USER: "mailer@example.com"
+  IMAP_SSL: "true"
+  IMAP_STARTTLS: "false"
```
yuanying: Please add `GITLAB_MONITORING_IP_WHITELIST: "10.0.0.0/8"`.
This network range should be pod network range. Default value "0.0.0.0/8" doesn't work.

israelshirk:
@solidnerd I have discontinued work on this due to the growing stability of Gitlab’s own helm package.  It does require some variation, but much less than this stuff - especially for running EE with things like kubernetes runners.  I’ll drop my manifests (based on helm’s dry-run exports) below.

PR #1213, opened 2017-05-14
Fixed changing name of a git user

nextsux:
I'm used to run as gitlab user (whole team has setup this user in remotes). There were few problems when GITLAB_USER is changed.


assets/runtime/env-defaults
```diff
@@ -21,6 +21,7 @@ else
 fi
 
 ## SSH
+GITLAB_SSH_USER=${GITLAB_SSH_USER:-$GITLAB_USER}
```
solidnerd: Default value for `$GITLAB_USER`is missing

assets/build/install.sh
```diff
@@ -34,7 +34,7 @@ paxctl -Cm `which nodejs`
 rm -rf /etc/ssh/ssh_host_*_key /etc/ssh/ssh_host_*_key.pub
 
 # add ${GITLAB_USER} user
-adduser --disabled-login --gecos 'GitLab' ${GITLAB_USER}
+adduser --disabled-login --gecos 'GitLab' ${GITLAB_USER} --home ${GITLAB_HOME}
```
solidnerd: Why is setting the home directory needed ?

assets/runtime/env-defaults
nextsux: It's defined in Dockerfile

assets/build/install.sh
nextsux: Without this adduser creates /home/$GITLAB_USER by default but everything else points to /home/$GITLAB_HOME

PR #1224, opened 2017-05-23
Upgrade to 9.2.0

andrewstuart:
Just a few changes were required here, and everything seems to be running straighforwardly for me.

The only thing that might cause issues is the fact that I had to add `bundle lock --update=omniauth-google-oauth2` because for some reason the Gemfile.lock in the 9.2 repo seems to have some version incompatibilities.


andrewstuart:
I also updated the install.sh script to pull the version numbers for downstream projects out of the gitlab-ce repo itself, since they're included as `*_VERSION` files.

phpb-com:
You will also need to compile gitlab-shell golang assets -  see https://gitlab.com/gitlab-org/gitlab-ce/blob/master/doc/update/9.1-to-9.2.md

The bundled fails due to bundler 1.15.0, see https://gitlab.com/gitlab-org/gitlab-ce/issues/32653

Nice PR, thanks!

andrewstuart:
Haha, that's exactly the issue I struggled with for an hour before adding the `gem lock --update=omniauth-google-oauth2`. I think that issue should be worked around in this PR.

Ilya-Kuchaev:
Any news here?

andrewstuart:
Yeah, not sure what we're waiting for. I've added a comment to the [gitlab issue](https://gitlab.com/gitlab-org/gitlab-ce/issues/32653) mentioned above about this specific workaround, but I don't think that will be fixed at the source for 9.2.0 since it's already been released.

As it is, the code in the PR builds and runs without any issues as far as I've seen over the last 24 hours.

Dockerfile
```diff
@@ -1,12 +1,9 @@
 FROM sameersbn/ubuntu:14.04.20170503
 MAINTAINER sameer@damagehead.com
 
-ENV GITLAB_VERSION=9.1.4 \
+ENV GITLAB_VERSION=9.2.0 \
     RUBY_VERSION=2.3 \
     GOLANG_VERSION=1.6.3 \
-    GITLAB_SHELL_VERSION=5.0.2 \
```
solidnerd: At the moment I don't want to have them dynamically in the `install.sh`. So please re add them there because it's easier to change the version and have an overview which version is currently used.

assets/build/install.sh
```diff
@@ -47,6 +47,14 @@ exec_as_git git config --global core.autocrlf input
 exec_as_git git config --global gc.auto 0
 exec_as_git git config --global repack.writeBitmaps true
 
+# shallow clone gitlab-ce
+echo "Cloning gitlab-ce v.${GITLAB_VERSION}..."
+exec_as_git git clone -q -b v${GITLAB_VERSION} --depth 1 ${GITLAB_CLONE_URL} ${GITLAB_INSTALL_DIR}
+
+GITLAB_SHELL_VERSION=$(cat ${GITLAB_INSTALL_DIR}/GITLAB_SHELL_VERSION)
```
solidnerd: We could do this. But first we should check if an VERSION has been set from the Dockerfile.

assets/build/install.sh
```diff
@@ -113,7 +117,9 @@ if [[ -d ${GEM_CACHE_DIR} ]]; then
   mv ${GEM_CACHE_DIR} ${GITLAB_INSTALL_DIR}/vendor/cache
   chown -R ${GITLAB_USER}: ${GITLAB_INSTALL_DIR}/vendor/cache
 fi
-exec_as_git bundle install -j$(nproc) --deployment --without development test aws
+
+exec_as_git bundle lock --update=omniauth-google-oauth2
```
solidnerd: This should be a separate commit. To make an easier revert if it's working correctly in the upstream again.

assets/build/install.sh
```diff
@@ -113,7 +117,9 @@ if [[ -d ${GEM_CACHE_DIR} ]]; then
...
+exec_as_git bundle install --deployment --without development test aws
```
solidnerd: Here you squash your last commit from the parallel build.

assets/build/install.sh
```diff
@@ -128,7 +134,7 @@ exec_as_git yarn install --production --pure-lockfile
 
 echo "Compiling assets. Please be patient, this could take a while..."
 #Adding webpack compile needed since 8.17
-exec_as_git bundle exec rake assets:clean assets:precompile webpack:compile USE_DB=false SKIP_STORAGE_VALIDATION=true >/dev/null 2>&1
+exec_as_git bundle exec rake assets:clean assets:precompile webpack:compile USE_DB=false SKIP_STORAGE_VALIDATION=true
```
solidnerd: Same as the comment before.

Ilya-Kuchaev:
[9.2.1](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/CHANGELOG.md#921-2017-05-23)

solidnerd:
Please Post only relevant information for the 9.2.0 release.

Am 24.05.2017 8:07 vorm. schrieb "Kuchaev Ilya" <notifications@github.com>:

Dockerfile
andrewstuart: Yeah, the only reason I added this is that it's a pain to Google the separate latest gitlab versions, and which ought to be used for which build.

This way, rebuilding the container for any GitLab version just involves changing a single variable and it'll pull the artifacts compatible with that version, per the included _VERSION files.

It's not a big deal for me -- I can add them back in and update to the latest; just thought it would ease your maintenance in the future.

solidnerd:
I would appreciate some last changes. Could you remove your changed Makefile from the `Update to 9.2.0` commit and make it as separate as well and squash the following commits as one.

- Upgrade to 9.2.0
- Update docs, readme, k8s, changelog
- Add manual env setup as dockerfile defaults

And remove the `Kick travis` commit. 

Also do check if the `.gitlab-ci.yml` has been changed since `9.1` ?

andrewstuart:
> Also do check if the .gitlab-ci.yml has been changed since 9.1 ?

Do you mean in the gitlab-ce repository? The local file seems to be pretty dead simple, and shouldn't require updates (since I bumped the version in `VERSION`).

phpb-com:
@andrewstuart I think @solidnerd means `assets/runtime/config/gitlabhq/gitlab.yml`. And there is one change that is missing from this PR.

```diff
    stuck_ci_jobs_worker:
      cron: "0 * * * *"
    # Execute scheduled triggers
--- trigger_schedule_worker:
+++ pipeline_schedule_worker:
      cron: "* */12 * * *"
    # Remove expired build artifacts
    expire_build_artifacts_worker:
```

andrewstuart:
@phpb-com Alright, the latest rebase should now include that change. :+1:

phpb-com:
@andrewstuart Awesome! 👍 

Btw, not too important yet, but there is an additional step for gitlab-shell, to compile golfing binaries. So far only hello-world is included, but that might change in the near future.

```
6. Update gitlab-shell

cd /home/git/gitlab-shell

sudo -u git -H git fetch --all --tags
sudo -u git -H git checkout v$(</home/git/gitlab/GITLAB_SHELL_VERSION)
sudo -u git -H bin/compile
```

solidnerd:
@phpb-com Thanks for the interpretting my typo correct. 

@andrewstuart I had one last thing. 
```bash

solidnerd:
Thanks for patient and work.

PR #1232, opened 2017-05-28
Add environment variable to configure pipeline schedule worker cron job

Glowbal:
- Adds a new environment variable to configure the cron notation for the Gitlab pipeline schedule worker introduced in Gitlab v9.2.0.
- The current value is `'0 */12 * * *'`, which I left as the default value.
- I have a couple of pipelines I run on a schedule and they have to run at different times. With the current setting, they will only run once every 12 hours (so twice a day); I need one to run at 04:00 and another at 05:00. Having the option to configure the cron notation for the schedule worker solves this.

Gitlab example config uses a different default value;
```
    # Execute scheduled triggers
    pipeline_schedule_worker:
      cron: "19 * * * *"
```
Should the default be switched to this?


solidnerd:
Hey, 
the function is missing to replace the cron the `gitlab.yml`. So In your case it would be only 
```

File: assets/runtime/env-defaults [hunk 1]
```diff
@@ -101,6 +101,9 @@ GITLAB_WORKHORSE_TIMEOUT=${GITLAB_WORKHORSE_TIMEOUT:-5m0s}
 GITLAB_ARTIFACTS_ENABLED=${GITLAB_ARTIFACTS_ENABLED:-true}
 GITLAB_ARTIFACTS_DIR="${GITLAB_ARTIFACTS_DIR:-$GITLAB_SHARED_DIR/artifacts}"
 
+## Cron Jobs
+GITLAB_PIPELINE_SCHEDULE_WORKER_CRON=${GITLAB_PIPELINE_SCHEDULE_WORKER_CRON:-'0 */12 * * *'}
```
splattael: The default value for `GITLAB_PIPELINE_SCHEDULE_WORKER_CRON` should read `19 * * * *`. See https://about.gitlab.com/gitlab-com/settings/#cron-jobs

solidnerd:
LGTM ! @splattael  I will add the new default behaviour

PR #1237, opened 2017-06-01
Bring registry nginx config on-par with others

fmauNeko:
Hello,

Here's a fix for allowing access to the registry behind a reverse proxy.
Like for the other services, if there's no SSL certificates, we just serve it over HTTP instead of HTTPS (Otherwise, we get an infinite redirection loop)

I based it on how it was done for the main service, and the Pages service.

Have a good day !

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,23 @@
+sudo: required
+language: bash
+services:
+  - docker
+env:
+  matrix:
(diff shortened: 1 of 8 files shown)


fmauNeko:
Just rebased it on latest master

fmauNeko:
As this PR is taking a long time, I pushed it myself to the Hub as fmauneko/gitlab.
Will be deleted as soon as this or #1259 are merged.

PR #1240, opened 2017-06-06
fix hash link

green-mike:
uppercase hash link is not active at chrome(version 58.0.3029.110)


solidnerd:
@green-mike  Thanks ;)

PR #1265, opened 2017-06-28
Upgrade to Gitlab 9.3.1

abeltay:
https://about.gitlab.com/2017/06/27/gitlab-9-dot-3-dot-1-released/


jostyee:
since 9.3 has been merged, this pr should be rebased and upgrade to the latest 9.3.3 version, which fixed a gitlab api major bug.

abeltay:
I've rebased for 9.3.1, I think @solidnerd may prefer to release versions in order.

@solidnerd shall I bump this to 9.3.1 or 9.3.3?

jostyee:
Both 9.3.1 and 9.3.2 are buggy, why bother tagging them?

abeltay:
Okay, I've updated it to 9.3.3

solidnerd:
@abeltay I prefer to have it in the release versions order. I would also like to have separate Pull Requests for a better overview of this.

@jostyee This could be important for some users. So that they have other issues that occures in the next patch version. When we bumping it up to the latest version we could not help this users. 

Another point is that I want personally to support the same gitlab version that are released by the upstream.

abeltay:
I've reverted the changes to 9.3.1

solidnerd:
@abeltay If you want you can prepare the other patches with PR's as well. I will gonna review them. When I back at home and probably release them today ;) .

abeltay:
Done! 😄

solidnerd:
I merged my version it was easier to resolve the conflics there but thanks for your work ;)

PR #1267, opened 2017-07-02
Change base image to ubuntu:trusty

abeltay:
@solidnerd would you prefer to use the official base image instead?


abeltay:
Closing since it doesn't seem like it will be used.

nixel2007:
There are so many useful unmerged PRs...

PR #1288, opened 2017-07-16
Regenerate secrets on container startup

MichaelEischer:
The gitlab container includes the files ${GITLAB_INSTALL_DIR}/.gitlab_{shell,workhorse}_secret .
However, these must be kept secret and should therefore not be included in the image.

I've found at least two possible attacks: The internal API intended only for use by the gitlab-shell can be accessed using the shell secret leading to information disclosure and bypass of 2FA.

Fixing this requires deleting both files during the container build process and recreating them on container startup. Although gitlab will also create both files automatically, there seems to be a race condition that from time to time breaks the gitlab-shell on the first startup of the container. The commit therefore calls `openssl rand` on first start to generate the secrects. The type of secret mirrors that from gitlab.


solidnerd:
LGTM !

PR #1296, opened 2017-07-20
Upgrade to GitLab 9.3.8

greyferret:
With R2 lib build dependencie
https://gitlab.com/gitlab-org/gitlab-ce/issues/35342


solidnerd:
Hey @greyferret,
I would use your PR instead of mine. For this it would be great to make two commits. First for the Upgrade of GitLab and a second one for the dependency. In future when we are upgrading to Ubuntu Xenial or higher than we don't need the compilation anymore. So than we can revert this commit with no effort.

solidnerd:
If you need help by this let me know.

solidnerd:
@greyferret  Why closing this ?

greyferret:
@solidnerd Sorry, i'm github newbie :) I made a new request with upgrade to 9.3.9 #1297

PR #1314, opened 2017-07-31
Use ARG to update GITLAB_VERSION var from VERSION file

untalpierre:
- Passing GITLAB_VERSION var from VERSION file in root folder.
- Update deprecated MAINTAINER for LABEL maintainer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ services:
   - docker
 
 script:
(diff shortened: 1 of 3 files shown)


klud1:
It all seems Travis CI doesn't use the makefile so the build arg needs to be passed to travis itself, just updated it

PR #1332, opened 2017-08-23
Uploading of Backups to OpenStack Swift

R2ZER0:
This merge request adds support for automatic uploading of backups to OpenStack Swift, in the same vein as the existing uploads to AWS or GCE.

Gitlab already supported OpenStack in the underlying config yml, so this isn't anything major, just exposing access to it via environment variables too.

I have successfully tested this using OVH's Cloud Object storage.


solidnerd:
Hey @R2ZER0, 
this sounds very interessting. I would like to merge but before I need to verify if it works correctly. Could make a little documentation. How you can setup a testing Swift store ?

R2ZER0:
Hi @solidnerd,

Sorry but I don't know how to actually set up an OpenStack Swift store! I use the cloud object storage from OVH, which is a Swift object store - afaik rackspace also offer something similar. I can recommend OVH for testing, given that it costs a couple of pence. Alternatively openstack seem to have DevStack, for quick setup on a local VM (but I have not tried this):
https://docs.openstack.org/devstack/latest/
https://ask.openstack.org/en/question/3298/how-to-install-devstack-for-testing-swift/

As for documentation, I've added the new env variables to the README, and added a usage section in the style of GCE/AWS.

GeorgeforChina:
Recently met a problem ，on the premise of not to install openstack, I want to install a multi server of swift, could you tell me whether can give help, the opentacke.com show  that to be based on openstack to install on the basis of.so if someone has installed  multi server of swift ,can you sharing you experience.thanks!

simon-jouet:
It's a bit of an old MR but would it be possible to get that merged in? having OpenStack Swift  backups enabled through the ENV like the rest would be great

PR #1333, opened 2017-08-24
Update to v9.5.1

lenovouser:
**[`gitlab.com/ce/tags#9.5.1`](https://gitlab.com/gitlab-org/gitlab-ce/tags/v9.5.1)**

PR #1334, opened 2017-08-24
[WIP] Upgrade GitLab to 9.5.1

solidnerd:


solidnerd:
I was to late #1333 appeard first . So in this case I will close it.

Issue #1347, opened 2017-09-08
Inconsistent performance due to frequent unicorn worker restarts


chaosversum:
I experienced some strange slow downs while navigating the gitlab web ui. While most of the requests are served in under 1 second, 20-30% of the requests needed 4-5 seconds to be served.

I digged a bit into it and found that the the unicorn worker restarts almost every 1-3 requests with just me navigating the web ui and 2 runners locking for jobs. I could even reproduced the problem on a fresh container with 9.5.3 by checking the unicorn.stderr.log while navigating the web ui.

Researching for a while gave me some more evidence that the frequent worker restarts might be the problem: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/2421

Unicorn workers while idle:

>   PID     ELAPSED   RSS COMMAND
>  583    02:36:41 327872 unicorn_rails master -c /home/git/gitlab/config/unicorn.rb -E production
>  671    02:36:19 385752 sidekiq 5.0.4 gitlab [0 of 25 busy]
> 6534       23:15 324424 unicorn_rails worker[1] -c /home/git/gitlab/config/unicorn.rb -E production
> 6548       23:03 324300 unicorn_rails worker[2] -c /home/git/gitlab/config/unicorn.rb -E production
> 6597       22:45 321052 unicorn_rails worker[0] -c /home/git/gitlab/config/unicorn.rb -E production

Unicorn workers after 3 web ui request (all restarted):

>  PID     ELAPSED   RSS COMMAND
>  583    02:38:46 328028 unicorn_rails master -c /home/git/gitlab/config/unicorn.rb -E production
>  671    02:38:24 385752 sidekiq 5.0.4 gitlab [0 of 25 busy]
> 7474       00:13 362032 unicorn_rails worker[1] -c /home/git/gitlab/config/unicorn.rb -E production
> 7488       00:08 352336 unicorn_rails worker[2] -c /home/git/gitlab/config/unicorn.rb -E production
> 7502       00:01 313600 unicorn_rails worker[0] -c /home/git/gitlab/config/unicorn.rb -E production

After some more research, i decided to increase the worker memory limit by building my own image from the source and adding the following lines to the unicorn.rb:

> ENV['GITLAB_UNICORN_MEMORY_MIN'] = "629145600"
> ENV['GITLAB_UNICORN_MEMORY_MAX'] = "734003200"

[...]

PR #1353, opened 2017-09-20
Increase default memory max to address 9.5.x performance problems

lekoder:
Gitlab defaults to unicorn memory limit of ~350MB. In 9.5.x this causes poor performance, as unicorn workers are constantly killed and re-spawned, leading to terrible performance even on small deployments. This problem is detailed in issue #1347 

While it is possible to adjust this limit with environment variables, this PR should provide a sane default.


phenomax:
Thanks for your PR!
Could you also add your new env. var to the [env-defaults](https://github.com/sameersbn/docker-gitlab/blob/master/assets/runtime/env-defaults) and the [functions](https://github.com/sameersbn/docker-gitlab/blob/master/assets/runtime/functions) class, so that it can be used in production?
Otherwise it won't be replaced/used

lekoder:
@phenomax `functions` seems to [source env-defaults already](https://github.com/sameersbn/docker-gitlab/blob/master/assets/runtime/functions#L3), so we should have this variable set up in production.

Since passing this option in environment already works (just without default), I'm not sure what change is required in `unicorn.rb`. Could you elaborate?

andypost:
btw Why 512M? min-max is 300-350 https://gitlab.com/gitlab-org/gitlab-ce/blob/master/config.ru#L10

andypost:
Also interesting measures https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/2421

lekoder:
I took 512M from @chaosversum at #1347. I have tested this on our installations and it worked great - performance was greatly increased, to levels we seen in 8.x. Whole docker uses now 3.8G of RAM (with 8 unicorn workers). With default of 350M workers are killed constantly. Issue itself is detailed in #1347.

phenomax:
@lekoder Correct me, if I am wrong, but don't we have to add the environment variable to the unicorn config and replace it to apply our custom value? Otherwise we won't change the unicorn memory in any way?

lekoder:
@phenomax Isn't env picked up by [config.ru](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/config.ru#L10) anyway? If it was not, setting it via docker command line would not work either. Then again, I didn't actually dwell that deep into gitlab, I don't have experience in ruby. But from my limited knowledge, if env is there, it should work.

phenomax:
Oh sorry. I didnt see the env var in the config.ru

lekoder:
...and my confusion came from fact, that it was you who pointed this particular line in #1347. Good thing that we have this cleared out.

solidnerd:
LGTM ! I will merge this for the 10.0 release and make a note in changelog for that.

PR #1378, opened 2017-10-11
Add helm chart

yuanying:
Helm chart is useful to install this project into Kubernetes
environment. This makes kubernetes manifest configurable for
each own environment.


solidnerd:
Hey @yuanying, 
this is looking really great. I think we should synchronise it with #1211 and see the progress for that. Probably what I think is that we make an new folder for kubernetes it could be `k8s` and it contains the chart and the plain from #1211

yuanying:
@solidnerd Hi, thanks for your review!
Yes, I saw #1211.
Which is better to create k8s folder before this commit or after this?
Or I should create k8s folder in this PR?

solidnerd:
@yuanying Is it still in WIP or could I do a final check to merge it ?

yuanying:
Hi, I'm still working this but I'll update soon.

yuanying:
@solidnerd Hi, sorry for late reply. Now it's ok to merge this!

PR #1428, opened 2017-11-30
Add support for building the enterprise edition

grimmy:
This PR adds support for building the enterprise edition as well.  This is controlled via a Docker build argument.  It defaults to building the community edition but you can build the enterprise edition with

```bash
docker build --build-arg GITLAB_EDITION=ee -t sameersbn/gitlab:10.2.2-ee .
```
I kept the documentation changes light as I figured you might want to tweak them a bit.

PR #1429, opened 2017-12-01
Add support for SSH keys in ldap

grimmy:
This PR just exposes a settings for the LDAP option `sync_ssh_keys`.


solidnerd:
Hey @grimmy, 
thanks for your work. I think it will be not merged because this a feature of gitlab-ee and not ce.

grimmy:
Sure, but this is just a config update that's ignored in the CE version.  But when this is paired up with #1428  it's actually quite useful.

PR #1443, opened 2017-12-19
Accelerate map_uidgid

kauden:
Hi,

I use the function "map_uidgid" and each restart of the container is at least 30 seconds for the command chown.
With this change I am 1 second.

I added the path ${GITLAB_HOME} at the end to not get an error if the "find" finds nothing.

does this change suit you?


jfreax:
Can we please merge this? This is super useful and speeds up updates a lot.

sameersbn:
thanks for the contribution

PR #1456, opened 2017-12-29
Improve docker-compose setup for local dev

drubin:
* Bump compose version to 2.2 to leverage cache_from
* Mount runtime assets as volume to allow easier editing
* Add basic how to

I thought I would document my experience working on this project so that others might have an easier time. I found it pretty hard to get started/debug things.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '2'
+version: '2.2'
 
 services:
   redis:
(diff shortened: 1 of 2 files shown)


drubin:
The `cache_from` is a nice trick to leverage dockers built-in caching from pre-built images. If you aren't keen on either the `cache_from` or the `asset/runtime` mounting let me know and will adjust it.

PR #1458, opened 2017-12-30
Allow running on PaX kernels

xen0n:
`ruby` needs the PaX MPROTECT flag disabled to run on PaX-enabled systems because of JIT. Mark it as such.


solidnerd:
Thanks for this contribution 

LGTM !

PR #1459, opened 2018-01-01
[WIP] Replace backup settings configuration to ERB syntax.

drubin:
This is a prototype to check if this is something you would be willing to experiment with? 

Gitlab uses settingsLogic that support ERB syntax in config files, as well as all rails configuration files such as `database.yaml`

See
https://gitlab.com/gitlab-org/gitlab-ce/blob/master/config/initializers/1_settings.rb#L5
https://github.com/settingslogic/settingslogic

> Settingslogic is a simple configuration / settings solution that uses an ERB enabled YAML file. It has been great for our apps, maybe you will enjoy it too. Settingslogic works with Rails, Sinatra, or any Ruby project.

This can be tested by simply booting up gitlab and testing nothing changes, also by starting the rails console and dumping the settings object. 

     bundle exec rails console production
     Settings.backup 

I know this diverges slightly from the standard `sameersbn/` images but gitlab's configuration files and manual sed hacks to get a proper templating language

Let me know if this is something worth exploring and I would be happy to work on converting it.

This would allow us to enable more complex configuration such as Redis Authentication, it simplifies the AWS/Google backup sections.

### DIFF
--- a/assets/runtime/config/gitlabhq/gitlab.yml
+++ b/assets/runtime/config/gitlabhq/gitlab.yml
@@ -502,30 +502,32 @@ production: &base
     pg_schema: {{GITLAB_BACKUP_PG_SCHEMA}}     # default: nil, it means that all schemas will be backed up
     upload:
       # Fog storage connection settings, see http://fog.io/storage/ .
-      #start-aws
+      <% if ENV['AWS_BACKUPS'] %>
       connection:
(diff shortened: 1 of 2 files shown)


drubin:
@sameersbn @solidnerd Would this be something of interest? Or should I close the PR

PR #1460, opened 2018-01-01
[WIP] Leverage docker's cache to speed up build times and CI feedback

drubin:


drubin:
Due to a bug in the way, the COPY of build time assets work the cache doesn't include the `bash install.sh` code which accounts for most of the build time so this is not as useful as it could be.

    COPY assets/build/ ${GITLAB_BUILD_DIR}/
    RUN bash ${GITLAB_BUILD_DIR}/install.sh

https://travis-ci.org/sameersbn/docker-gitlab/builds/323818604#L404 

We could play around with changing the docker file to, but I am not sure its worth the effort.

    COPY assets/build/install.sh ${GITLAB_BUILD_DIR}/
    RUN bash ${GITLAB_BUILD_DIR}/install.sh

Issue #1427, opened 2017-11-30
port 80 


lrjxhchina:
how can i modify the port 80  because this 80 is used

balajidl:
Are you starting docker using docker command or docker-compose ?

If docker command, see step 3 at https://github.com/sameersbn/docker-gitlab#installation.
If via docker-compose, you can change it at https://github.com/sameersbn/docker-gitlab/blob/master/docker-compose.yml#L30

p-v-a:
it seems that GITLAB_PORT and GITLAB_SSH_PORT are ignored. I can see /home/git/gitlab/config/gitlab.yml contains correct ports, while actual nginx configuration still refers to port 80.

Once I changed templates in /etc/docker-gitlab/runtime/config/nginx file the nginx switched to a port specified in GITLAB_PORT.

GITLAB_SSH_PORT more complex, there is no provisioning in /etc/docker-gitlab/runtime/functions to modify sshd config.

PR #1462, opened 2018-01-05
Add GITLAB_PORT into nginx templates

p-v-a:
This fixes an issue when GITLAB_PORT value are ignored and Ngnix keeps listen on http port 80. This is address https://github.com/sameersbn/docker-gitlab/issues/1427


p-v-a:
I just realized that it's not actually required, it works as it is. So closing this request.

PR #1463, opened 2018-01-05
Add GITLAB_BACKUP_STRATEGY option

schizo99:
We were having issues with the backups
Setting STRATEGY=copy solved our problem.
Since this is supported in the original gitlab I think it's good to have as an environment variable instead of having to manually modifying the container each time it's created.

[Backup strategy option](https://docs.gitlab.com/ee/raketasks/backup_restore.html#backup-strategy-option)


drubin:
Thanks for adding this.

Could you please add these variables to `env-defaults` file with a default blank value, as with all other ENV vars. 

It seems that setting it to blank should keep the "default" behaviour. (which should already be the default)

https://github.com/gitlabhq/gitlabhq/blob/bbba5eb06a09c94556cfc20c5580b3f779c49fea/lib/backup/files.rb#L20

schizo99:
Other variables were not defined in env-defaults and since it's blank by default I was hesitant in adding it.

I've tested the rake command with STRATEGY=<blank> without issues.

README.md
```diff
@@ -857,6 +857,7 @@ Below is the complete list of available options that can be used to customize yo
 | `GITLAB_BACKUP_ARCHIVE_PERMISSIONS` | Sets the permissions of the backup archives. Defaults to `0600`. [See](http://doc.gitlab.com/ce/raketasks/backup_restore.html#backup-archive-permissions) |
 | `GITLAB_BACKUP_TIME` | Set a time for the automatic backups in `HH:MM` format. Defaults to `04:00`. |
 | `GITLAB_BACKUP_SKIP` | Specified sections are skipped by the backups. Defaults to empty, i.e. `lfs,uploads`. [See](http://doc.gitlab.com/ce/raketasks/backup_restore.html#create-a-backup-of-the-gitlab-system) |
+| `GITLAB_BACKUP_STRATEGY` | Set to `copy` to avoid `file changed as we read it` errors. [See](https://docs.gitlab.com/ee/raketasks/backup_restore.html#backup-strategy-option) |
```
solidnerd: Here should be the link to the  documentation for the community edition and not enterprise edition replace this with this `https://docs.gitlab.com/ce/raketasks/backup_restore.html#backup-strategy-option`

schizo99:
When can I expect this pull request to be merged?
As it is our backups are failing because of this.

PR #1464, opened 2018-01-05
Update change log for 10.3.3 release

drubin:
@solidnerd You seemed to miss the changelog entry for the 10.3.3 release so I updated it for you.

Would you mind also including the pax mark in the release notes? https://github.com/sameersbn/docker-gitlab/releases/tag/10.3.3

Thanks again for including my fix so quickly.


solidnerd:
Thanks for this I think we should gonna try to add this for new releases in the format to link to the pr.

PR #1467, opened 2018-01-15
Improve Container registry docs

drubin:
Having gone through this process I thought I would update the documentation so it's correct and maybe a bit simpler without duplicating the docs which are already somewhere else.

* Fix commands to generate certs.
* Simplify docs by linking to official docs and having a basic quickstart

I also removed the Nginx example since it already had 
 
> We assume that you already have Nginx installed on your host system

### DIFF
--- a/docs/container_registry.md
+++ b/docs/container_registry.md
@@ -3,27 +3,33 @@ GitLab Container Registry
 Since `8.8.0` GitLab introduces a container registry. GitLab is helping to authenticate the user against the registry and proxy it via Nginx. By [Registry](https://docs.docker.com/registry) we mean the registry from docker whereas *Container Registry* is the feature in GitLab.
 
 - [Prerequisites](#prerequisites)
+    - [Assumptions](#assumptions)
 - [Installation](#installation)
+    - [Create Auth Tokens](#create-auth-tokens)
 - [Configuration](#configuration)
+    - [Available Parameters](#available-parameters)
+    - [ Storage driver](#storage-driver)
 - [Maintenance](#maintenance)
-    - [Creating Backups](#creating-backups)
-    - [Restoring Backups](#restoring-backups)
-- [Upgrading from an existing GitLab instance](#Upgrading-from-an-existing-GitLab-instance)
 
 # Prerequisites
 
   - [Docker Distribution](https://github.com/docker/distribution) >= 2.4
   - [Docker GitLab](https://github.com/sameersbn/docker-gitlab) >= 8.8.5-1
 
+**Before you start this guide you should have a general understanding of**
 
-# Installation
+ - [Docker Registry](https://docs.docker.com/registry/)
+ - [Docker Registry configuration options](https://docs.docker.com/registry/configuration)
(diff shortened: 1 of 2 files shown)


docs/container_registry.md
```diff
@@ -256,131 +183,15 @@ storage:
     - REGISTRY_STORAGE_S3_ACCESSKEY=AKIAKIAKI
     - REGISTRY_STORAGE_S3_SECRETKEY=secret123
     - REGISTRY_STORAGE_S3_BUCKET=gitlab-registry-bucket-AKIAKIAKI
-    - REGISTRY_CACHE_BLOBDESCRIPTOR=inmemory
+    - REGISTRY_STORAGE_CACHE_BLOBDESCRIPTOR=inmemory
```
drubin: This param was actually wrong and it crashes.

docs/container_registry.md
```diff
@@ -103,103 +120,17 @@ Then update the `volumes` and `environment` sections of your `gitlab` container:
...
-    }
-}
-
-server {
-    listen *:80;
-    server_name  registry.example.com;
-    server_tokens off; ## Don't show the nginx version number, a security best practice
-    return 301 https://$http_host:$request_uri;
-}
-```
```
mikehaertl: @drubin I think at least the crucial parts of the Nginx configuration should remain. E.g. which `proxy_set_header` you should have and the `proxy_pass`.

I've actually added the full config to leave no questions about how the setup works. So I'd say, we should keep all of it.

docs/container_registry.md
```diff
@@ -103,103 +120,17 @@ Then update the `volumes` and `environment` sections of your `gitlab` container:
...
-gitlab:
-    environment:
-    - GITLAB_REGISTRY_ENABLED=true
-    - GITLAB_REGISTRY_HOST=registry.gitlab.example.com
-    - GITLAB_REGISTRY_API_URL=http://registry:5000
-    - GITLAB_REGISTRY_KEY_PATH=/certs/registry-auth.key
-    - GITLAB_REGISTRY_ISSUER=gitlab-issuer
-...
-```
+Here is an example of all configuration parameters that can be used in the GitLab container. See the [Available Configuration Parameters](https://github.com/sameersbn/docker-gitlab#available-configuration-parameters).
```
mikehaertl: If you remove those, then we should also remove the sentence "Here is an example of all configuration parameters ...". They are no longer here. But I agree that those could probably be removed.

docs/container_registry.md
drubin: The reason I removed them is the entire setup requires working proxy server setup and it. The assumption nd the docs don't match

> We assume that you already have Nginx installed on your host system

I would  rather add an additional doc on setting up a proxy for gitlab and the registry and link to that. What do you think?

mikehaertl: Not sure what you mean here:

> We assume that you already have Nginx installed on your system

Yes, Nginx is installed. And it may already be configured for the GitLab site (without the registry). But not the registry. That's why I included the configuration that you have to add to your existing Nginx installation.

Maybe we could change that sentence to:

> We assume that you already have Nginx installed and use it as reverse proxy for GitLab. We add another site configuration for the Container registry.

PR #1468, opened 2018-01-15
Create version bump script

drubin:
Simple script to replace the version of gitlab in all the files.
Works on osx and linux

### DIFF
--- a/bump-version.sh
+++ b/bump-version.sh
@@ -1,0 +1,19 @@
+#!/bin/bash -e
+# Replaces the current gitlab version with a new version
+
+display_usage() { 
+	echo "Replaces the current gitlab version" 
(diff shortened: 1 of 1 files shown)


pieterlange:
This also rewrites the `Changelog.md` which is probably unintended.

I just clumsily did something similar while staging a version bump (which should be *easy*) but it is easily manually patched. Excuse my french:
`for file in $(grep -r 10.3.3 . | cut -d: -f1 | sort | uniq); do sed -i 's/10.3.3/10.3.4/g' $file ;done`

drubin:
Thanks. Will fix it to exclude the changelog. But yes it should be easier. The annoying bit is also making it work both on mac and Linux.

drubin:
FIxed to exclude the changelog.

drubin:
@solidnerd ping?

PR #1477, opened 2018-01-19
Add support for building the enterprise edition

grimmy:
This is a repost of PR 1428 on it's own branch this time.

This PR adds support for building the enterprise edition as well. This is controlled via a Docker build argument. It defaults to building the community edition but you can build the enterprise edition with

```
docker build --build-arg GITLAB_EDITION=ee -t sameersbn/gitlab:10.2.2-ee .
```

I kept the documentation changes light as I figured you might want to tweak them a bit.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,7 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv E1DD270288B4E60
  && rm -rf /var/lib/apt/lists/*
 
 COPY assets/build/ ${GITLAB_BUILD_DIR}/
+ARG GITLAB_EDITION=ce
 RUN bash ${GITLAB_BUILD_DIR}/install.sh
 
 COPY assets/runtime/ ${GITLAB_RUNTIME_DIR}/
--- a/README.md
+++ b/README.md
@@ -135,10 +135,18 @@ You can also pull the `latest` tag which is built from the repository *HEAD*
 docker pull sameersbn/gitlab:latest
 ```
 
-Alternatively you can build the image locally.
+Alternatively you can build the images locally.
+
+## Community Edition
 
 ```bash
 docker build -t sameersbn/gitlab github.com/sameersbn/docker-gitlab
+```
+
+## Enterprise Edition
+
+```bash
+docker build --build-arg GITLAB_EDITION=ee -t sameersbn/gitlab-ee github.com/sameersbn/docker-gitlab
 ```
 
 # Quick Start
--- a/assets/build/install.sh
+++ b/assets/build/install.sh
@@ -1,7 +1,14 @@
 #!/bin/bash
 set -e
 
-GITLAB_CLONE_URL=https://gitlab.com/gitlab-org/gitlab-ce.git
+GITLAB_EDITION=${GITLAB_EDITION:-ce}
+
+# if we're using the enterprise edition suffix the version with -ee
+if [ x"${GITLAB_EDITION}" = x"ee" ] ; then
+	GITLAB_VERSION="${GITLAB_VERSION}-ee"
+fi
+
+GITLAB_CLONE_URL=https://gitlab.com/gitlab-org/gitlab-${GITLAB_EDITION}.git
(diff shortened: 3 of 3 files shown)


lordvlad:
would love to see this merged

snics:
Yeah, would it be nice if someone could comment on that?

martinspaeth:
I would like to give ee a try using this patch. Do one know if the container ce and ee are compatible? I want to be able to switch back to ce if ee din’t Match our expectations...

Thank you for your help

tomislater:
We have been using it for three months without a hitch.

martinspaeth:
Great! Thanks for your fast response  @tomislater. Two more questions. If I run the EE container do I need to provide a license key directly or does it run with the ce features until I provide the corresponding key? And what Happens if I use it as EE and the key expires.

tomislater:
https://docs.gitlab.com/ee/user/admin_area/license.html

reynold-lariza:
Hi @grimmy and everyone, I was able to build a container using this patch -- but when I tried this on to of existing 11.8.1, Gitlab stuck on the temp page (the page when its still just starting up), also workhorse and gitaly failed to start. Was able to revert back to CE, but unable to migrate to EE :(

martinspaeth:
Is anyone here running ee containerized with those images lately? I would love to switch to ee but I don’t want to give up the current docker setup...

DZDomi:
We are currently running an EE gitlab instance (11.6) with this image and also this PR. Everything is working great so far. I only had to change 1 setting in the nginx configuration cause we are using Gitlab Pages with custom domain names and TLS. But besides that all features are working as intended.

Maghin:
Hi,

I've been running Gitlab-EE 11.3.2 for 6 month with this image. You just have to replace your CE container with a new one created from this image. Nothing more to do. This works perfectly except that I lost the user SSH keys but I think it's not related.

I just want to warning about the license model Gitlab-EE uses : you have to pay a license for every "active user". An active user is in fact an "active account", no matter if he has already connected to Gitlab or not. So before migrating be sure to block every non used account.
On our little business it means we can no longer use automatic/bot user, LDAP authentication, or create a temporary access to a client to manage an issues (I can not explain my boss that he as to pay an annual license for a 5 minutes usage...).

Also good to know : Gitlab does not answer to questions about the "starter license" (probably not profitable enough).

It's a bit disappointing and I'm now going back to CE...

elitvinov:
@grimmy, can you please merge upstream and resolve conflicts.
Your patch to use Gitlab EE edition seem to be still working

Dockerfile
```diff
@@ -50,6 +50,7 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv E1DD270288B4E60
  && rm -rf /var/lib/apt/lists/*
 
 COPY assets/build/ ${GITLAB_BUILD_DIR}/
+ARG GITLAB_EDITION=ce
```
splattael: Since GitLab moved to a single codebase (see https://about.gitlab.com/blog/2019/08/23/a-single-codebase-for-gitlab-community-and-enterprise-edition/) the naming has changed (although the redirects still work!). It's not `gitlab-org/gitlab-foss` (suffix is `-foss`; former `-ce`) and `gitlab-org/gitlab` (no suffix; former `-ee`).

FYI removing the EE folder from `gitlab-org/gitlab` (`rm -fr ./ee`) gives you basically the FOSS version.

elitvinov: Correct, both EE => CE and CE => EE procedures require some manual work:
1. https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/downgrade_ee_to_ce/README.md
2. https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/update/upgrading_from_ce_to_ee.md

Some or procedures are not applicable to docker environments as they are applied at container start or build (cache clearing, deps installation, db migrations). So migrations are also the discussion point for test and documenting using sameersbn docker images.

I mean now conflict resolving is mainly use the new url for CE, as it made in master at 7ca29dfce0cedf51ff47da24c813c9b9984dddbc

PR #1488, opened 2018-01-27
Update base image to xenial

DoubleMalt:


solidnerd:
Hey,
it's already in there see #1479 . It will be available with the 10.4.0 release. So I think we can gonna close this.

PR #1514, opened 2018-02-19
Internal certificate for Docker Registry autogeneration, Docker Compose for Docker Swarm mode example and deployment guide

tiangolo:
This PR covers several features at the same time, I needed all of them, and as it was not obvious to me and took me some time, I thought it was worth sharing.

All the features are in the same PR because, at least for my use case, they were tied together.

Features:

* **Traefik** integration, handling TLS termination, load balancing, **HTTPS** certificate generation and renewal with **Let's encrypt**. Instead of needing to create a complex custom Nginx proxy on top of the Docker Registry, with several potential complications, it's all handled by Traefik.
* The methodology described makes the cluster compatible with more future stacks deployed with their independent HTTPS certificates.

It was heavily based on my experience creating this guide: https://github.com/senseta-os/senseta-base-project/blob/master/docker-swarm-cluster-deploy.md

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,7 +56,9 @@ RUN bash ${GITLAB_BUILD_DIR}/install.sh
 
 COPY assets/runtime/ ${GITLAB_RUNTIME_DIR}/
 COPY entrypoint.sh /sbin/entrypoint.sh
+COPY generate-registry-certificates.sh /sbin/generate-registry-certificates.sh
 RUN chmod 755 /sbin/entrypoint.sh
+RUN chmod 755 /sbin/generate-registry-certificates.sh
 
 EXPOSE 22/tcp 80/tcp 443/tcp
 
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@
     - [Shell Access](#shell-access)
 - [Features](#features)
  - [Container Registry](docs/container_registry.md)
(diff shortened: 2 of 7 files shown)


tiangolo:
I would like to know, have you been able to review the changes?

I would absolutely prefer to have the changes here than in a fork. Or at least to hear the maintainers and see if there are any concerns or objections.

If you haven't had the time but still plan to review it, I understand (it's happening to me on some projects), but I would love to hear from you...

tiangolo:
Guys @sameersbn , @solidnerd , do you think these changes would not be a good fit to this project? I understand if that's the case but please let me know.

I'm thinking of possibly creating an independent project that is based on this, and these images, that implement my changes as an additional layer on top. But of course, that would be more cumbersome than having it all here.

solidnerd:
Hey @tiangolo, 

sorry for no response so far. I think it was going under my radar. 

It would be cool that you rebase your branch. So that the new CI system can be triggered. So that we see if it's breaks something.

I really like the approach with traefik. I think we provide it as an alternative for the nginx deployment.

I think in can review this in the next 2 days and give your an equal review to your commendable pull request.

tiangolo:
Sorry for the delay @solidnerd ! I re-started the conversation and you responded right away... and then I got busy building [**FastAPI**](https://github.com/tiangolo/fastapi).

I just created a new PR to keep things clean. Starting from a fresh `master`, adding all these changes and updating all the versions.

Let's continue the conversation there. I'll close this PR.

PR #1530, opened 2018-02-28
Remove build files after installation finished

chengweiv5:
Signed-off-by: Chengwei Yang <yangchengwei@qiyi.com>


adamlc:
Because this is being done in an addition RUN command it'll actually create an additional layer, so won't really have any affect to the build size, assuming thats what you want to do!

chengweiv5:
@adamlc oh, that's correct! so it's should be removed in the install script.

chengweiv5:
updated

jostyee:
@solidnerd this pr can reduce almost 600MB of the final image size.

solidnerd:
But I think it will brake some functionality. So one case would be that `GITLAB_RELATIVE_URL_ROOT` is working also after this.

PR #1548, opened 2018-03-19
Update gitlab-rc.yml

fforootd:
If we use the default uploads dir path, it resides within the images location which is not great.

I guess an easy fix would be to supply the `GITLAB_UPLOADS_STORAGE_PATH`variable to point to the dir which is mounted to a persistent type of storage.


sameersbn:
> If we use the default uploads dir path, it resides within the images location which is not great.

@fforootd by default the `uploads` directory is located at `${GITLAB_DATA_DIR}/uploads` (via symlinks).

fforootd:
we do not require this anymore 😄

PR #1588, opened 2018-04-24
added GITLAB_LOG_ROTATE_FREQUENCY and GITLAB_LOG_ROTATE_RETENTION

hardikns:
I realized there is no way to update the logrotate frequency and retention parameters set by the build script. 
Here is a small MR that would help in setting that up.


solidnerd:
Hey, 

I'll check it out later or tomorrow (saturday).

evenh:
Have you had the time to have a look @solidnerd? :-)

Issue #1626, opened 2018-06-04
after docker-compose restart gitlab exited with error 


flashermaster:
Hi,
After a successful docker-compose up, I'd like to reboot the container. 
I tried docker-compose down/stop, then I'd like to start it again I received  errors like this:
```
/volume1/docker/gitlab# docker-compose up
Starting gitlab_postgresql_1 ... 
Starting gitlab_redis_1 ... 
Starting gitlab_postgresql_1
Starting gitlab_redis_1 ... done
Creating gitlab_gitlab_1 ... 
Creating gitlab_gitlab_1 ... done
Attaching to gitlab_postgresql_1, gitlab_redis_1, gitlab_gitlab_1
postgresql_1  | WARNING: no logs are available with the 'db' log driver
redis_1       | WARNING: no logs are available with the 'db' log driver
gitlab_1      | WARNING: no logs are available with the 'db' log driver
gitlab_gitlab_1 exited with code 143
gitlab_1      | WARNING: no logs are available with the 'db' log driver
gitlab_gitlab_1 exited with code 143
......
```
I watched the log and the last unusual log is: 
/sbin/entrypoint.sh: line 22:   558 Terminated              /usr/bin/supervisord -nc /etc/supervisor/supervisord.conf

Don't understand the meaning.
My yml file is just the standard info and the first up works well. 

Could someone help me with this issue?
Thank you so much.

adamlc:
I'm getting a similar error too after an upgrade 

```
mysql: [Warning] Using a password on the command line interface can be insecure.
/sbin/entrypoint.sh: line 22:   657 Terminated              /usr/bin/supervisord -nc /etc/supervisor/supervisord.conf
```

flashermaster:
@adamlc There is only endless reboot. I have to remove the container and also the data in my volume to make next clean docker-compose up work.

adamlc:
I've managed to get it working by commenting out the database migration stuff in entrypoint.sh on line 22 and rebuilding the docker image.

Obviously this isn't good because there may well be migrations in 10.8.3. But at least its running now! If there are migrations I doubt it'll be anything major in a point release.

flashermaster:
@adamlc Thanks.Would you please give me any suggestion of which former version of sameersbn/gitlab would be pretty stable?

adamlc:
@flashermaster well as of this morning I had been running 10.8.1 until I tried to upgrade 😞

boy28213:
[...]

PR #1628, opened 2018-06-04
Correct entrypoint.sh causing bootloop in some cases

adferrand:
Following issue #1626 

Entrypoint executes following commands:
* kill -15 $SUPERVISOR_PID
* wait $SUPERVISOR_PID

This may fail if process is killed sufficiently fast: indeed process may already be killed when wait is invoked, and then results into an error. Command `ps -p` ensures process is still here before waiting to its state to change.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -19,7 +19,7 @@ case ${1} in
         SUPERVISOR_PID=$!
         migrate_database
         kill -15 $SUPERVISOR_PID
-        ps h -p $SUPERVISOR_PID && wait $SUPERVISOR_PID
+        ps h -p $SUPERVISOR_PID > /dev/null && wait $SUPERVISOR_PID
         rm -rf /var/run/supervisor.sock
         exec /usr/bin/supervisord -nc /etc/supervisor/supervisord.conf
         ;;
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -19,7 +19,7 @@ case ${1} in
         SUPERVISOR_PID=$!
         migrate_database
         kill -15 $SUPERVISOR_PID
-        wait $SUPERVISOR_PID
+        ps h -p $SUPERVISOR_PID && wait $SUPERVISOR_PID
         rm -rf /var/run/supervisor.sock
         exec /usr/bin/supervisord -nc /etc/supervisor/supervisord.conf
         ;;


solidnerd:
@adferrand Seems working for me. I tested it with a restart after a migration.

solidnerd:
Are there any things left or can we start with a new release ?

adferrand:
@solidnerd And I deployed it on my production environment without any issue. I think we are good to go.

solidnerd:
Okay I will start with this than now. Thanks for your work 👍

paddy-hack:
IIUC, you've tried to address a race between the `kill` and `wait` invocations, but didn't you just replace that with a race between the `ps` and `wait` invocations?  
If so, would putting the `wait` in the background before the `kill` and then foregrounding it solve that?  Like so
```sh
wait $SUPERVISOR_PID &
kill -15 $SUPERVISOR_PID
fg
```
Not sure if `fg` can be used in the container though.

If using `bash`'s `wait` the `-n` option might be useful to detect that `$SUPERVISOR_PID` no longer exists.

paddy-hack:
About the empirical evidence, I just had both 10.8.3 and 10.8.3-1 loop on a clean startup :unamused:.  I use `docker-compose` and all the containers use a bind mounted volume (with a shared parent directory).  That is to say, they use something like `./test/$containername`.  The clean startup is one where the shared parent directory does not exist yet.

As for your theoretical analysis, I think it depends a lot on the implementation details of the shell that's in use, i.c. `/bin/bash`.  As for my example code, well, that's completely untested but I do put the `wait` in the background for a reason :smile:

adferrand:
And so died my hypothesis. As a scientist, I have to admit my mistake ^^ I will remember that shells are definitely programming languages, and that everything must be designed with the multi-threading nature of the underlying os in mind. 

So your solution is better as it ensures that no race can occur. I will make a new PR. 

Thanks!

adferrand:
Hum. In fact executing `wait $SUPERVISOR_PID &` leads to a "wait pid is not a child of this shell". Indeed executing the wait in background makes it be executed in a sub-child of the current shell. Then it is not allowed to watch the process of pid `$SUPERVISOR_PID`, which is also a sub-child of the current shell.

Started from the foreground, wait, as a built-in command, stays in the shell and then can act normally.

So of course we could make a while loop on the /proc/$SUPERVISOR_PID until it disapear, but here goes again the race between while and its execution statement !

@paddy-hack, do you have an idea ?

adferrand:
Or will just do a `wait $SUPERVISOR_PID || true` to trap the potential error on wait, and we are good to go. Anyway if wait fails, it means that the process is already dead, and that is what we want before continuing.

PR #1637, opened 2018-06-12
Fix for LFS Large File upload

directrix1:
This tells nginx to not buffer proxied requests. This is required to avoid "io errors" on large LFS transfers.


GJRTimmer:
@directrix1 Quick question; I have no problem with the current image with LFS transfers of roughly 700+ MB. Can you provide an information how large your files are ?

PR #1663, opened 2018-07-09
fix base image tag

sameersbn:
the base image was not tagged properly. fixed now


sameersbn:
ci failed because the image has not been built by dockerhub yet.

PR #1665, opened 2018-07-14
use file_env to have the ability to read secrets from orchestrator-pr…

earzur:
…ovided secret files (rancher 1.x)

Hello, this PR updates the entrypoint to add:
* the file_env() function and  that we can find in many database-related container
* ability to read variables that we don't want to commit into stack descriptors 

we are using rancher 1.6 here, and this makes our life much easier ;-)

Hope you'll find it useful.


drubin:
You can already load via files, not 100% like you specified but you could always drop an `env` file in `/gitlab-configs`

See https://github.com/sameersbn/docker-gitlab/blob/master/assets/runtime/functions#L4

PR #1671, opened 2018-07-19
Ed25519 GPG key import fix

EmbeddedAndroid:
This issue was raised in this thread: https://gitlab.com/gitlab-org/gitlab-ce/issues/36845,
but only ever fixed in omnibus-gitlab. I was able to reproduce this issue with the v11.0.1
release of docker-gitlab. The fix was to add gnupg2 as runtime dep, and now Ed25519 GPG
keys import fine.

Signed-off-by: Tyler Baker <forcedinductionz@gmail.com>


solidnerd:
LGTM !

PR #1678, opened 2018-07-27
clean up the $HOME/.cache directory after installation

sameersbn:
Yarn stores every package in a global cache in your user directory on
the file system. This ends up to be around ~400M of extra utilized by
the image and has no benefits whatsoever.

Additionally `go` populates a build cache with will also be cleared
when the ~/.cache directory is purged

/cc @solidnerd

### DIFF
--- a/assets/build/install.sh
+++ b/assets/build/install.sh
@@ -386,3 +386,6 @@ EOF
 # purge build dependencies and cleanup apt
 DEBIAN_FRONTEND=noninteractive apt-get purge -y --auto-remove ${BUILD_DEPENDENCIES}
 rm -rf /var/lib/apt/lists/*
+
+# clean up caches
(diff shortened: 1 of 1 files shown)


sameersbn:
Oh.. good point! let me check

sameersbn:
the changes do not appear to have any effect on the `relative_url` functionality. I think we're good.

solidnerd:
LGTM than !

paddy-hack:
> the changes do not appear to have any effect on the relative_url functionality

Alas, you spoke too soon :cry:
The long story is in #1696, but if your instance is behind a proxy and not given proxy info at *run-time* you're hosed.  I've been banging my head against the wall for two days before I figured out what was going on :unamused:

I do believe that an instance should not require HTTP(S) access to stuff on the other side of the proxy at *run-time*.

PR #1705, opened 2018-08-19
integrate with circleci

sameersbn:


.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
+version: 2
+jobs:
+  build:
+    working_directory: /workdir
+    docker:
+      - image: docker:17.06.0-ce-git
```
solidnerd: we could specify here `docker:stable-git` than we don't need to worry about the docker version.

.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
...
+        environment:
+          IMAGE_NAME: "sameersbn/gitlab"
+
+    steps:
+      - checkout
+
+      - setup_remote_docker:
+          version: 17.06.0-ce
```
solidnerd: Here than `docker:stable`

solidnerd:
Otherwise LGTM !

.circleci/config.yml
sameersbn: circle ci explicitly lists the supported docker versions in https://circleci.com/docs/2.0/building-docker-images/#docker-version. I'll update it to `18.05.0-ce` and give it a test. But as you can tell using the `stable` tag could have unintended consequences

solidnerd:
Should we remove travis than ?

sameersbn:
> Should we remove travis than ?
yep, added a commit the removes travis.yml

PR #1712, opened 2018-09-01
Upgrade GitLab to 11.2.1

solidnerd:


insekticid:
@solidnerd why not 11.2.3?

solidnerd:
@insekticid I will release every version of 11.2.X to be consistent

PR #1714, opened 2018-09-01
Upgrade GitLab to 11.2.2

solidnerd:


solidnerd:
You need only to wait that the build ist finished. The Progress can be viewed on https://hub.docker.com/r/sameersbn/gitlab/builds/

PR #1727, opened 2018-09-22
Upgrade GitLab to 11.3.0

solidnerd:


solidnerd:
Important is that this will be my last release before my vacation. I'll be back on 3.10.2018. 

@sameersbn Could you handle the next releases until I will be back.

solidnerd:
LGTM !

PR #1734, opened 2018-10-04
Upgrade GitLab to 11.3.3

solidnerd:


solidnerd:
So there are no changes of gitlab in this release. We'll skip this than.

solidnerd:
My assumption was not correct. This should be the correct fix. So we take this than also.

PR #1757, opened 2018-11-01
Upgrade GitLab CE to 11.4.4 and golang to 1.10.4

drujd:
decided not to upgrade golang to 1.11.1 as only 1.9 and 1.10 is tested for gitlab-workhorse, even though golang should be backwards compatible


solidnerd:
LGTM !

Issue #1765, opened 2018-11-07
chown of ${GITLAB_BACKUP_DIR} can fail on NFS


lewismarshall:
When using a network backup NFS device for backups, startup will fail when trying to `chown` the backups directory as specified here:
https://github.com/sameersbn/docker-gitlab/blob/1f71b9999a51c4075ba9468ea0b584e0a6f4eb28/assets/runtime/functions#L1332

It would be desirable to be able to configure this as an optional action (when the storage NFS device is NOT configurable).

solidnerd:
I will close it if your problems still persists. Please feel free to reopen your issue.

PR #1766, opened 2018-11-07
Make the chown of backup files optional

lewismarshall:
Fixes issue #1765


assets/runtime/env-defaults
```diff
@@ -5,6 +5,7 @@ DEBUG=${DEBUG:-$DEBUG_ENTRYPOINT}
 ## GITLAB CORE
 GITLAB_TEMP_DIR="${GITLAB_DATA_DIR}/tmp"
 GITLAB_BACKUP_DIR="${GITLAB_BACKUP_DIR:-$GITLAB_DATA_DIR/backups}"
+GITLAB_BACKUP_DIR_CHOWN=true
```
solidnerd: This doesn't allow to overwrite it. I think you should do
```bash
GITLAB_BACKUP_DIR_CHOWN=${GITLAB_BACKUP_DIR_CHOWN:-true}
```

lewismarshall: @solidnerd indeed thanks good spot - pushed the fix

PR #1768, opened 2018-11-08
Fix container timezone

gjrtimmer:
***Gitlab can have a different timezone in relation to the container***

# Introduction
Default the containers run with timezone set to `UTC`, which means that the cron daemon can be using a different timezone than gitlab, because the timezone from gitlab can be configured with `GITLAB_TIMEZONE`.

# Use case
My current TimeZone is `Europe/Amsterdam` Gitlab has the corrent time, but because the container time is `UTC` it means that the cron daemon which runs the backups does this ` hour to early.

# Provided solution

* ADDED 
  * `configure_container_timezone()` -> `assets/runtime/functions`
* MODIFY
  * Added `configure_container_timezone()` to function `configure_system()` before the function `install_configuration_templates` because timezone is a lower level then templates.
  * README
    * ADD: Variable `TZ`
    * UPD: Variable `GITLAB_TIMZONE`

## configure_container_timezone()

The provided function has a sanity check which will use `UTC` as default. The sanity check will check that the user provided value is a valid timezone, by checking if the provided timezone exists within the `tzdata` zoneinfo.

# Configuration

Container TimeZone will use the environment variable `TZ`, and defaults to `UTC`. The variable `TZ` expects values in Canonical format. Link to acceptable values has been added to the  README. The format which is expected causes that the `GITLAB_TIMEZONE` cannot be used here because `GITLAB_TIMEZONE` can accept timezone in different formats.


GJRTimmer:
@solidnerd Request review

splattael:
@GJRTimmer Thanks for the PR!

Would mounting `/etc/localtime` into the container have the same effect? E.g. `-v /etc/localtime:/etc/localtime:ro`?

GJRTimmer:
@splattael Never tested that, and I don't know if it would be a solution. I think there would be quite a few situations where there is a complete different timezone between the containers and the host. With this fix it the container will get the correct timezone. Got this fix from `docker` on how to set the container timezone correctly by using the `TZ` variable.

This image might even be used by parties that are hosting gitlab environments for clients in different timezone. So I don't want to mount the localtime from the host I want to set it correctly.

GJRTimmer:
@solidnerd @sameersbn Any Update

GJRTimmer:
Rebased on new master

GJRTimmer:
@solidnerd Rebased on new master

GJRTimmer:
@solidnerd Rebased to 11.5.0

GJRTimmer:
Rebased to 11.5.1

GJRTimmer:
@solidnerd rebased

GJRTimmer:
@solidnerd Rebased

drubin:
@GJRTimmer Hi is there a reason you don't just set the `TZ` env variable?

assets/runtime/functions
```diff
@@ -1448,12 +1448,30 @@ generate_ssh_host_keys() {
   chmod 0644 ${GITLAB_DATA_DIR}/ssh/*.pub
 }
 
+configure_container_timezone() {
+  TIMEZONE=${TZ:-UTC}
```
drubin: This default config I think should be set in `env-defaults` if it is in fact required.

sameersbn: +1

sameersbn:
sorry about the delay in attending to this PR. I really have no issues getting it merged but I am just curious to know if there is a use-case that just setting the `TZ` env variable is not suffice?

GJRTimmer:
> sorry about the delay in attending to this PR. I really have no issues getting it merged but I am just curious to know if there is a use-case that just setting the `TZ` env variable is not suffice?

Sorry for the delay; I have environment variable TZ=Europe/Amsterdam set on the container. Still the shell is not reporting the current correct time which results that for example my backup files have an offset of 2 hours because of GMT+2 (Daylight Saving Time).

Proposed PR would fix that.

assets/runtime/functions
GJRTimmer: PR Updated;

GJRTimmer:
@solidnerd @sameersbn Rebased & Implemented conversation suggestion

GJRTimmer:
@solidnerd @sameersbn Why is the build not completing ?

sameersbn:
needs a rebase.

gjrtimmer:
@sameersbn Rebased, very sorry for the delay

sameersbn (approved):

PR #1775, opened 2018-11-19
which -s flag is not universal

lordvlad:
google cloud platform shell, for example, does not support the -s flag.


solidnerd:
LGTM !

PR #1776, opened 2018-11-20
fix double CI job

GJRTimmer:
Exclude master branch from docker:build:branches

@solidnerd Currently when building master is run twice, both the jobs, `docker:build` and `docker:build:branches` are triggerd when building the master.

Provided fixes this.
Think we both missed it.


solidnerd:
LGTM !

PR #1791, opened 2018-11-27
Fix: Gitlab Pages Custom Domain Setup

GJRTimmer:
When running the Gitlab Pages within a `custom domain setup` according to the documentation of Gitlab. The container will fail to start `nginx` due to the following error:

```bash
3051#3051: duplicate listen options for [::]:443 in /etc/nginx/sites-enabled/gitlab-pages:26
```

This fix allows the user to disable the gitlab pages proxy when providing their own HTTP Load Balancer for Gitlab Pages.

@solidnerd Request Review.


assets/runtime/env-defaults
```diff
@@ -414,6 +414,7 @@ GITLAB_PAGES_HTTPS=${GITLAB_PAGES_HTTPS:-false}
 GITLAB_PAGES_EXTERNAL_HTTP=${GITLAB_PAGES_EXTERNAL_HTTP:-}
 GITLAB_PAGES_EXTERNAL_HTTPS=${GITLAB_PAGES_EXTERNAL_HTTPS:-}
 GITLAB_PAGES_ACCESS_CONTROL=${GITLAB_PAGES_ACCESS_CONTROL:-false}
+GITLAB_PAGES_NGINX_DISABLE=${GITLAB_PAGES_NGINX_DISABLE:-false}
```
solidnerd: I don't like the naming of this variable, because I think the for the user it's not directly clear what this variable can do. I think more clear way would be `NGINX_GITLAB_PAGES_PROXY` or `GITLAB_PAGES_NGINX_PROXY`

GJRTimmer: Changes Done.

Variable renamed: `GITLAB_PAGES_NGINX_PROXY` changed, default, from `false` to `true` and updated all corresponding functions so that the variable makes sense.

`GITLAB_PAGES_NGINX_PROXY` Default: `true`

When set to `false` it will disable the proxy, README has been updated

@solidnerd Request Check

GJRTimmer:
@solidnerd Can you restart the Circle-CI. Don't know what caused the failure

GJRTimmer:
Rebased to 11.5.1

GJRTimmer:
@solidnerd Rebased

solidnerd:
Hey,

this will be released with `11.6.0`. Thanks for your work. I need to cherry-pick it manually but you can find it here https://github.com/sameersbn/docker-gitlab/commit/5c59e16edb53976330dbcacfd5e22b265b780b2c .

PR #1746, opened 2018-10-19
Fix LDAP username issue

frol2103:
* this fixes the issue related to https://gitlab.com/gitlab-org/gitlab-ce/issues/43926#note_73018296

Fixes #1745

I am not sure of the procedure to add an array in a env variable but I didn't find any example.

Please do not hesitate to tell if change needs to be made.

PR #699, opened 2016-05-25
add env vars for LDAP Attributes

hknochi:
added the possibility to set the attributes for LDAP Users with environment variables

PR #1814, opened 2018-12-14
Add support for custom LDAP name value in config

jodlajodla:
Proposed changes add support for custom setting of `name` property in `ldap` part of `gitlab.yml` config.


GJRTimmer:
Possible duplicate of #1746
Possible duplicate of #699

PR #1824, opened 2018-12-31
Add GITLAB_IMPERSONATION_ENABLED

pbe-axelor:
Add `GITLAB_IMPERSONATION_ENABLED` config to disable admin impersonation


solidnerd:
LGTM ! 

I think it can be in the next release this means it would be in `11.6.1` available.

PR #1826, opened 2019-01-02
Traefik Gitlab v11.6.0

tiangolo:
Continuation of #1514 

---

This PR covers several features at the same time, I needed all of them, and as it was not obvious to me and took me some time, I thought it was worth sharing.

All the features are in the same PR because, at least for my use case, they were tied together.

Features:

* **Traefik** integration, handling TLS termination, load balancing, **HTTPS** certificate generation and renewal with **Let's encrypt**. Instead of needing to create a complex custom Nginx proxy on top of the Docker Registry, with several potential complications, it's all handled by Traefik.
* The methodology described makes the cluster compatible with more future stacks deployed with their independent HTTPS certificates.

It was heavily based on my experience creating this guide: https://github.com/senseta-os/senseta-base-project/blob/master/docker-swarm-cluster-deploy.md


docker-compose.swarm-stack.yml
```diff
@@ -0,0 +1,215 @@
...
+    ports:
+    - "22:22"
+    volumes:
+    - gitlab-gitlab:/home/git/data:Z
+    - gitlab-certs:/certs
+    deploy:
+      labels:
+        - "traefik.frontend.rule=Host:git.example.com"
+        - "traefik.enable=true"
+        - "traefik.port=80"
```
solidnerd: Instead of going through the nginx you can call the workhorse directly with `traefik.port=8181`

docker-compose.swarm-stack.yml
```diff
@@ -0,0 +1,215 @@
...
+    - redis
+    - postgresql
+    ports:
+    - "22:22"
+    volumes:
+    - gitlab-gitlab:/home/git/data:Z
+    - gitlab-certs:/certs
+    deploy:
+      labels:
+        - "traefik.frontend.rule=Host:git.example.com"
```
solidnerd: For an easier management it would be more nice that use an variable to get the consistency other multiple values.

This means it would look like this
```
 - "traefik.frontend.rule=Host:${GITLAB_HOST:-git.example.com}"
```

This can be now overriden in two different ways. 

1. Pass it directly with command `GITLAB_HOST=2-git.example.com` docker stack deploy -c docker-compose.swarm-stack.yml
2. You can provide an file called `.env` and write the environment variables directly
```bash
#.env
GITLAB_HOST=2-git.example.com
```

docker-compose.swarm-stack.yml
```diff
@@ -0,0 +1,215 @@
...
+        # Traefik service that listens to HTTPS
+        - "traefik.webservice.frontend.entryPoints=https"
+    networks:
+      - traefik-public
+      - default
+    environment:
+    - DEBUG=false
+
+    - GITLAB_REGISTRY_ENABLED=true
+    - GITLAB_REGISTRY_HOST=registry.example.com
```
solidnerd: It would also recommend to use an env here:

```
 GITLAB_REGISTRY_HOST=${GITLAB_REGISTRY_HOST:-registry.example.com}
```

docker-compose.swarm-stack.yml
```diff
@@ -0,0 +1,215 @@
...
+    - DB_PASS=password
+    - DB_NAME=gitlabhq_production
+
+    - REDIS_HOST=redis
+    - REDIS_PORT=6379
+
+    - GITLAB_HTTPS=true
+    - SSL_SELF_SIGNED=false
+
+    - GITLAB_HOST=git.example.com
```
solidnerd: - GITLAB_HOST=${GITLAB_HOST:-git.example.com}

docs/container_registry.md
```diff
@@ -42,15 +42,16 @@ lifetime like this:
 ```bash
 mkdir certs
 cd certs
-openssl req -new -newkey rsa:4096 > registry.csr
-openssl rsa -in privkey.pem -out registry.key
+# Generate a random password password_file used in the next commands
+openssl rand -hex -out password_file 32
```
solidnerd: Why do we need a password file to generate the key with a password ?

docs/docker-swarm-traefik-registry.md
```diff
@@ -0,0 +1,553 @@
...
+docker service create \
+    --name traefik \
+    --constraint=node.role==manager \
+    --publish 80:80 \
+    --publish 8080:8080 \
+    --publish 443:443 \
+    --mount type=bind,source=/var/run/docker.sock,target=/var/run/docker.sock \
+    --mount type=volume,source=traefik-public-certificates,target=/certificates \
+    --network traefik-public \
+    traefik:v1.5 \
```
solidnerd: It can be `traefik:v1.7`

docs/docker-swarm-traefik-registry.md
```diff
@@ -0,0 +1,553 @@
...
+* Create an environment variable with your email, to be used for the generation of Let's Encrypt certificates:
+
+```bash
+export EMAIL=admin@example.com
+```
+
+* Create a Traefik service:
+
+```bash
+docker service create \
```
solidnerd: In general instead of a long command. I would also like to see a separate `docker-compose.traefik.stack.yml`. It's better to maintain instead of a command.

entrypoint.sh
```diff
@@ -2,8 +2,13 @@
 set -e
 set -o pipefail
 
+if [[ $GITLAB_REGISTRY_GENERATE_INTERNAL_CERTIFICATES == 'true' ]]; then
```
solidnerd: It would be more accurate to move this under `assets/runtime/function` and add the env to `
/assets/runtime/env-defaults`
1. Create an own function for your certificate generation 
2. Provide the env in in env-defaults

Also to be backwards compatible it should set to `false` by default.

docs/container_registry.md
tiangolo: Because otherwise, `openssl` asks for a "passphrase" interactively. So it can't be automatized.

That's just to be consistent with the script `generate-registry-certificates.sh` that does it all automatically when the env var `GITLAB_REGISTRY_GENERATE_INTERNAL_CERTIFICATES` is set to `'true'`.

With the commands as they are, they can be copy-pasted directly, without adding any additional input (no passphrase).

docs/docker-swarm-traefik-registry.md
tiangolo: Good catch!

docker-compose.swarm-stack.yml
tiangolo: Awesome.

I've been using that recently but imagined you wouldn't like having extra files for this.

docker-compose.swarm-stack.yml
tiangolo: Done :heavy_check_mark:

docker-compose.swarm-stack.yml
tiangolo: Done :heavy_check_mark:

docker-compose.swarm-stack.yml
tiangolo: Done :heavy_check_mark:

docs/docker-swarm-traefik-registry.md
tiangolo: Sure. I updated it to have an external extra stack file for Traefik. And I updated the docs accordingly.

I also updated those same docs with the latest changes/updates/improvements from [my original post](https://github.com/tiangolo/medium-posts/tree/master/docker-swarm-mode-and-traefik-for-a-https-cluster).

entrypoint.sh
tiangolo: Perfect. Done. :heavy_check_mark:

tiangolo:
@solidnerd Thanks for the thorough code review. I think I already implemented all the requested changes.

If you see anything missing, please let me know.

tiangolo:
Hey @solidnerd , have you been able to check this? Is there something else I should change first?

tiangolo:
I know you must be busy @solidnerd, but here's a gently (and annoying) reminder to check this out when you get a chance.

I would like to add this GitLab repo to https://dockerswarm.rocks/ after this PR :smile:

solidnerd:
I think we can merge this as well. WDYT @sameersbn  ?

sameersbn:
sgtm. @tiangolo could you do a rebase with the master so that we're sure things are working after the merge

tiangolo:
Thanks @sameersbn and @solidnerd !

I just did, I sent a new PR with the updates in #2162 . It updates, simplifies, and supersedes this PR.

Feel free to close it when you read this message/feel appropriate.

solidnerd:
I'll gonna close this now through the merge of #2162.

PR #1852, opened 2019-02-01
Upgrade to 11.7.3

greyferret:


solidnerd:
Hey, 

I think we should ship `11.7.1` and `11.7.2` because they weren't released while they got problems in the QA.

greyferret:
@solidnerd Ok, I deleted 11.7.1 and 11.7.2

leslau:
Can someone merge this plz, thats some security issues that's fixed in this version!

pieterlange:
11.7.1 and 11.7.2 were never released to the public upstream either so i guess we can consider those internal project numbers? Just :shipit: ?

PR #1884, opened 2019-03-24
Add support for GITLAB_SHELL_CUSTOM_HOOKS_DIR

taomaree:
[Custom Git Hooks](https://docs.gitlab.com/ce/administration/custom_hooks.html) introduced in GitLab Shell 4.1.0 and GitLab 8.15. 

Default custom git hooks dir `/home/git/gitlab-shell/hooks` can be overwrite by upgrade gitlab version, it's better we can set hook dir. ENV GITLAB_SHELL_CUSTOM_HOOKS_DIR works like "custom_hooks_dir" in gitlab.rb.

Tests pass with gitlab 11.8.0


taomaree:
global custom hook is useful for administrator.
@solidnerd  please review it :)

solidnerd:
Hey @taomaree ,
I think our default label needs to be changed to something that it's not looking like an action. It should be something like `stale` or `rotten`. We enabled this only to get rid of not active issues because this project is a little bit older now these days and I hope from this that it will help us to see which issues still persist or need help from us by now.

So I got a question. Do you still use this feature?

sameersbn (approved):

PR #1906, opened 2019-04-22
Prefer CI_REGISTRY_PASSWORD over CI_REGISTRY_PASS

splattael:
Since 9.0 GitLab defines `CI_REGISTRY_PASSWORD` (which equals to `CI_JOB_TOKEN`) - see [predefined variables](https://docs.gitlab.com/ee/ci/variables/predefined_variables.html#variables-reference).

Using `CI_REGISTRY_PASS` is confusing and we should prefer the usage of `CI_REGISTRY_PASSWORD`. However, in order to be backwards-compatible we'll still read the deprecated `CI_REGISTRY_PASS`.


solidnerd:
This makes sense to me. I think a better reuse is better instead of providing an extra variable.

LGTM !

PR #1928, opened 2019-05-31
Fix nginx config template gitlab-pages-ssl

greyferret:
ipv6only parameter can only be set once and turned on by default. If you set it >1 time, nginx will fail at startup.

https://nginx.org/en/docs/http/ngx_http_core_module.html#listen

> ipv6only=on|off
this parameter (0.7.42) determines (via the IPV6_V6ONLY socket option) whether an IPv6 socket listening on a wildcard address [::] will accept only IPv6 connections or both IPv6 and IPv4 connections. This parameter is turned on by default. It can only be set once on start.

### DIFF
--- a/assets/runtime/config/nginx/gitlab-pages-ssl
+++ b/assets/runtime/config/nginx/gitlab-pages-ssl
@@ -23,7 +23,7 @@ server {
 ## Pages serving host
 server {
   listen 0.0.0.0:443 ssl;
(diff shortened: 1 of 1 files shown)


greyferret:
Rebased to upstream master.

@solidnerd review, please :)

solidnerd:
LGTM !

PR #1944, opened 2019-07-03
Add support for AWS S3 signature version

bartoszcisek:
Some 3rd party storage solutions allow upload via S3 protocol but still
use v2 of signature schema. [Fog library used by gitlab](https://docs.gitlab.com/ee/administration/uploads.html#s3-compatible-connection-settings)
allows to override default version 4.


bartoszcisek:
Hey! Thanks for a review. I provided a default signature version `4`  as [fog library used in gitlab](https://docs.gitlab.com/ee/administration/uploads.html#s3-compatible-connection-settings) has been using `4` as a default. Before this PR signature couldn't be overridden in an easy way, so probably the majority of users were keeping default `4`. Additionally, [AWS S3 is not using signature version](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingAWSSDK.html#specify-signature-version) `2` anymore. Only some third-party vendors like mine [QNAP OSS](https://www.qnap.com/en-us/how-to/tutorial/article/hosting-s3-and-openstack-compatible-object-storage-services-in-qts/) still use deprecated version `2`. 

This was my reasoning for providing default as `4`.  If default `2` would be more suitable, I'll prepare patch right away. What do you think?

solidnerd:
Hey @bartoszcisek,
I checked your pr and it was working for me. Thanks for your patient. We'll include this in the  `12.0.0` release.

Issue #1962, opened 2019-07-25
Compilation error


peterwillcn:
```
 node ./scripts/frontend/postinstall.js
success Dependency postinstall check passed.
Done in 10.56s.
Compiling assets. Please be patient, this could take a while...
warning Resolution field "ts-jest@24.0.0" is incompatible with requested version "ts-jest@^23.10.5"
`yarn:check` finished in 5.020868173 seconds
Missing Rails.application.secrets.secret_key_base for production environment. The secret will be generated and stored in config/secrets.yml.
Missing Rails.application.secrets.otp_key_base for production environment. The secret will be generated and stored in config/secrets.yml.
Missing Rails.application.secrets.db_key_base for production environment. The secret will be generated and stored in config/secrets.yml.
Missing Rails.application.secrets.openid_connect_signing_key for production environment. The secret will be generated and stored in config/secrets.yml.
rake aborted!
PostgreSQL is the only supported database from GitLab 12.1
/home/git/gitlab/config/initializers/1_postgresql_only.rb:3:in `<top (required)>'
/home/git/gitlab/vendor/bundle/ruby/2.5.0/gems/activesupport-5.2.3/lib/active_support/dependencies.rb:285:in `load'
/home/git/gitlab/vendor/bundle/ruby/2.5.0/gems/activesupport-5.2.3/lib/active_support/dependencies.rb:285:in `block in load'
/home/git/gitlab/vendor/bundle/ruby/2.5.0/gems/activesupport-5.2.3/lib/active_support/dependencies.rb:257:in `load_dependency'
/home/git/gitlab/vendor/bundle/ruby/2.5.0/gems/activesupport-5.2.3/lib/active_support/dependencies.rb:285:in `load'
/home/git/gitlab/vendor/bundle/ruby/2.5.0/gems/railties-5.2.3/lib/rails/engine.rb:657:in `block in load_config_initializer'
/home/git/gitlab/vendor/bundle/ruby/2.5.0/gems/activesupport-5.2.3/lib/active_support/notifications.rb:170:in `instrument'
Compiling assets. Please be patient, this could take a while...
warning Resolution field "ts-jest@24.0.0" is incompatible with requested version "ts-jest@^23.10.5"
`yarn:check` finished in 5.020868173 seconds
Missing Rails.application.secrets.secret_key_base for production environment. The secret will be generated and stored in config/secrets.yml.
[...]

PR #1966, opened 2019-07-27
(try to fix issue #1962 Compilation error)

gnuhub:
#1962


gnuhub:
@solidnerd thank you very much!

solidnerd:
I think it can be closed since `12.1.x` it's fixed and the latest version can be build.

PR #1989, opened 2019-08-27
Fix HostKey function for ubuntu:bionic

### LINKED ISSUE #1987: Upgrade 12.2.0 (from 12.1.6) required host key generation
Maybe related to switch to `bionic`?

Just migrated from 12.1.6 to 12.2.0 image and the web UI worked fine but the ssh connections all failed.

Looking at the SSHD logs, there were lots of:

```No supported key exchange algorithms [preauth]
Could not load host key: /etc/ssh/ssh_host_rsa_key
Could not load host key: /etc/ssh/ssh_host_ecdsa_key
Could not load host key: /etc/ssh/ssh_host_ed25519_key
```

(these files did not exist)

I was able to resolve the issue with `ssh-keygen -A` to generate new host keys. This was a manual step and hasn't been required in previous upgrades.

solidnerd:
Closes #1987

### DIFF
--- a/assets/runtime/functions
+++ b/assets/runtime/functions
@@ -1439,7 +1439,7 @@ generate_ssh_key() {
 }
 
 generate_ssh_host_keys() {
-  sed -i "s|HostKey /etc/ssh/|HostKey ${GITLAB_DATA_DIR}/ssh/|g" /etc/ssh/sshd_config
+  sed -i "s|#HostKey /etc/ssh/|HostKey ${GITLAB_DATA_DIR}/ssh/|g" /etc/ssh/sshd_config
   if [[ ! -e ${GITLAB_DATA_DIR}/ssh/ssh_host_rsa_key ]]; then
     echo -n "Generating OpenSSH host keys... "
     generate_ssh_key rsa      ${GITLAB_DATA_DIR}/ssh/ssh_host_rsa_key


solidnerd:
Hey,
to make it a little bit more clear. During the startup of the container the original `/etc/ssh/sshd_config` will be changed. So in the original key stands `#HostKey /etc/ssh/...` the generate function will overwrite this to the `$GITLAB_DATA_DIR` in this case it's `/home/git/data` . The `/etc/ssh/sshd_config` isn't persisted outside of the container. Everytime a container is restarted or better recreated this change will happen through the runtime scripts. 

The recreation will happen through docker-compose because you will need to update your image to `12.2.1-1`. 

Here is an example output:

* 12.2.1

```
root@3b5720b892bc:/home/git/gitlab# cat /etc/ssh/sshd_config 
#	$OpenBSD: sshd_config,v 1.101 2017/03/14 07:19:07 djm Exp $

# This is the sshd server system-wide configuration file.  See
# sshd_config(5) for more information.

# This sshd was compiled with PATH=/usr/bin:/bin:/usr/sbin:/sbin

# The strategy used for options in the default sshd_config shipped with
# OpenSSH is to specify options with their default value where
# possible, but leave them commented.  Uncommented options override the
# default value.

#Port 22
#AddressFamily any
#ListenAddress 0.0.0.0
#ListenAddress ::

#HostKey /home/git/data/ssh/ssh_host_rsa_key
#HostKey /home/git/data/ssh/ssh_host_ecdsa_key
#HostKey /home/git/data/ssh/ssh_host_ed25519_key
```

* 12.2.1-1 (pre-release)

```
#	$OpenBSD: sshd_config,v 1.101 2017/03/14 07:19:07 djm Exp $

# This is the sshd server system-wide configuration file.  See
# sshd_config(5) for more information.

# This sshd was compiled with PATH=/usr/bin:/bin:/usr/sbin:/sbin

# The strategy used for options in the default sshd_config shipped with
# OpenSSH is to specify options with their default value where
# possible, but leave them commented.  Uncommented options override the
# default value.

#Port 22
#AddressFamily any
#ListenAddress 0.0.0.0
#ListenAddress ::

HostKey /home/git/data/ssh/ssh_host_rsa_key
HostKey /home/git/data/ssh/ssh_host_ecdsa_key
HostKey /home/git/data/ssh/ssh_host_ed25519_key
```

PR #1993, opened 2019-08-31
Add Sentry configuration to gitlab.yml

sue445:
c.f. https://gitlab.com/gitlab-org/gitlab-ce/blob/v12.2.3/config/gitlab.yml.example#L436-441


sue445:
Is there anything I can do for this patch?

PR #1994, opened 2019-09-01
Add prepared_statements to database.yml

sue445:
I want to use `docker-gitlab` with [PgBouncer](https://pgbouncer.github.io/).

But need to disable `prepared_statements`

c.f. 

* https://guides.rubyonrails.org/configuring.html#configuring-a-postgresql-database
* https://gitlab.com/gitlab-org/omnibus-gitlab/blob/d31e78538d82cea4ba9c40f79cac033f7cf84c9d/files/gitlab-cookbooks/gitlab/templates/default/database.yml.erb#L25

So I allowed pass this parameter  to `database.yml`


sue445:
Is there anything I can do for this patch?

assets/runtime/env-defaults
```diff
@@ -44,6 +44,7 @@ DB_NAME=${DB_NAME:-}
 DB_USER=${DB_USER:-}
 DB_PASS=${DB_PASS:-}
 DB_POOL=${DB_POOL:-10}
+DB_PREPARED_STATEMENTS=${DB_PREPARED_STATEMENTS:-}
```
solidnerd: I think the default case should be true. This would ensure that we got the old behaviour by default.

solidnerd:
```suggestion
DB_PREPARED_STATEMENTS=${DB_PREPARED_STATEMENTS:-true}
```

PR #2009, opened 2019-09-24
Add S/MIME Email Signing

pbe-axelor:
See https://docs.gitlab.com/ce/administration/smime_signing_email.html

Signed-off-by: Pierre Belloy <p.belloy@axelor.com>


assets/runtime/config/gitlabhq/gitlab.yml
```diff
@@ -70,6 +70,15 @@ production: &base
     email_display_name: {{GITLAB_EMAIL_DISPLAY_NAME}}
     email_reply_to: {{GITLAB_EMAIL_REPLY_TO}}
     email_subject_suffix: '{{GITLAB_EMAIL_SUBJECT_SUFFIX}}'
+    email_smime:
```
solidnerd: I think we need also a logic here to remove those values if `GITLAB_EMAIL_SMIME_ENABLE` is false. Check for the logic https://github.com/sameersbn/docker-gitlab/blob/master/assets/runtime/functions#L765 and https://github.com/sameersbn/docker-gitlab/blob/master/assets/runtime/config/gitlabhq/gitlab.yml#L633 for the yaml setup. This will remove the lines if it's not set and should the default behaviour.

pbe-axelor: Added `email-smime` block. The s/mime email configurations is now removed if `GITLAB_EMAIL_SMIME_ENABLE` is false

pbe-axelor:
Rebased onto upstream/master

assets/runtime/config/gitlabhq/gitlab.yml
```diff
@@ -70,6 +70,17 @@ production: &base
...
```
solidnerd:
```suggestion
    #start-email-smime
    email_smime:
```

assets/runtime/config/gitlabhq/gitlab.yml
```diff
@@ -70,6 +70,17 @@ production: &base
...
+      #start-email-smime
```
solidnerd:
```suggestion

```

pbe-axelor: `#start-email-smime` moved on top of `email_smime`

PR #1961, opened 2019-07-24
[WIP] Upgrade GitLab to 12.1.0

solidnerd:

PR #2031, opened 2019-10-22
Remove remaining references to MySQL

splattael:
MySQL support was removed in https://github.com/sameersbn/docker-gitlab/pull/1961

### DIFF
--- a/README.md
+++ b/README.md
@@ -929,7 +929,7 @@ Below is the complete list of available options that can be used to customize yo
 | `DB_ADAPTER` | The database type. Currently only postgresql is supported. Over 12.1 postgres force. Possible values: `postgresql`. Defaults to `postgresql`. |
 | `DB_ENCODING` | The database encoding. For `DB_ADAPTER` values `postresql` this parameter defaults and `utf8` respectively. |
 | `DB_HOST` | The database server hostname. Defaults to `localhost`. |
-| `DB_PORT` | The database server port. Defaults to `3306` for mysql and `5432` for postgresql. |
+| `DB_PORT` | The database server port. Defaults to `5432` for postgresql. |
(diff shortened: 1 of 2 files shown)


assets/build/install.sh
```diff
@@ -153,7 +153,7 @@ if [[ -d ${GEM_CACHE_DIR} ]]; then
   chown -R ${GITLAB_USER}: ${GITLAB_INSTALL_DIR}/vendor/cache
 fi
 
-exec_as_git bundle install -j"$(nproc)" --deployment --without development test mysql aws
+exec_as_git bundle install -j"$(nproc)" --deployment --without development test
```
splattael: Not sure about this. Do we still need `mysql` and `aws`? :thinking: 

Refs https://github.com/sameersbn/docker-gitlab/pull/1961/files#r337386468

solidnerd: You are right we could also drop `aws`and `mysql` according to https://docs.gitlab.com/ce/install/installation.html#install-gems . Mysql should be though I think it got missed during the rebasing in #1961 .

solidnerd: But to have it as a correctly behaviour you should have passed it in the `--without` option so this would install `aws`and `mysql`. So it needs to be like this:

```suggestion
exec_as_git bundle install -j"$(nproc)" --deployment --without development test aws mysql
```

splattael: I agree :+1: Let's the removal of `aws` and `mysql` in a separate PR :+1:

splattael: Tracked via https://github.com/sameersbn/docker-gitlab/issues/2035

splattael:
@solidnerd Rebased against latest `master` also reverted the `aws`/`mysql` change. We'll tackle this in https://github.com/sameersbn/docker-gitlab/issues/2035 :+1:

PR #2059, opened 2019-12-10
Add SSL_PROTOCOLS variable

toadstule:
ssl_protocols for Nginx are hard-coded at 'TLSv1 TLSv1.1 TLSv1.2'.  TLSv1 and TLSv1.1 are no longer considered secure.  Rather than changing the defaults, I thought it best to add an environment variable "SSL_PROTOCOLS" that will allow the end user to override the default setting. This change is similar to a past change that added the SSL_CIPHERS variable.


drujd:
Nice, but there it is also hardcoded in gitlab-pages-ssl and gitlab-registry

drujd:
Also, I strongly believe that a reasonable default is TLSv1.2 & TLSv1.3 and ciphers according to https://ssl-config.mozilla.org/#server=nginx&server-version=1.16.1&config=intermediate&openssl-version=1.1.1

Current defaults are really not secure anymore and make no sense since moving to bionic base

PR #2087, opened 2020-02-17
Docker pull appropriate/curl before Testing image

dotneB:


solidnerd:
Hey, 

thanks for your pull request. Why is an extra pull needed? This will help me to understand.

dotneB:
I was under the impression that `docker run` needed the image to be pulled prior to running it. But this is not the case and doesn't fix the problem I was trying to fix (which is the circleci sometime failing). I was a bit quick on the trigger for that PR, I'll close it a open a new one that does fix the issue, now that I've investigated more.

PR #2093, opened 2020-03-02
Allow modifying the sshd listen port

KashifSaadat:
Add the ability to modify the default listen port for SSHD on container startup. The default behaviour (listening on Port 22) is kept, and can be overridden by setting the following environment variable:
```
GITLAB_SSH_LISTEN_PORT: 2222
```

When this is set, `GITLAB_SSH_PORT` is respectively updated to match (if not explicitly set).


sameersbn:
lgtm

PR #2140, opened 2020-05-06
handle new gitaly logs correctly

achretien:
The current config added to gitlay config file is not complete.

I uncommented the [logging] group in the toml config file.

I also added logrotate config and user right to log folder


sameersbn:
Thanks for the fix

PR #2151, opened 2020-05-21
Fix accidental replacement if no backup is configured

coolersport:
If none of AWS or GCS backups are configured, sed command would accidentially replace the second "upload:" occurrence until the end of file.


solidnerd:
Hey @coolersport,
thanks for your PR. It looks very promising it tried already with default values and the ci is also passing. I need only to test the `aws` or `gcs` part than we are ready to go to ship it.

#2147

solidnerd:
LGTM !

PR #2162, opened 2020-05-26
✨ Add automatic shared self-signed certificates for Docker Registry, Docker Swarm, Traefik, and HTTPS

tiangolo:
✨ Add automatic shared self-signed certificates for Docker Registry, Docker Swarm, Traefik, and HTTPS.

This updates, simplifies, and supersedes #1826 .

I rebased from `master`, simplified the guide, and upgraded Traefik to v2.

---

It has:

* **Traefik** integration, handling TLS termination, load balancing, **HTTPS** certificate generation and renewal with **Let's encrypt**. Instead of needing to create a complex custom Nginx proxy on top of the Docker Registry, with several potential complications, it's all handled by Traefik. Just upgraded to use Traefik version 2 :sparkles: 
* The methodology described makes the cluster compatible with other stacks deployed with their independent HTTPS certificates.

---

Some parts of the guide in the previous PR were a bit old, so I simplified it to reduce the maintenance burden and instead point to [DockerSwarm.rocks](https://dockerswarm.rocks) for the generic parts (deploying a Docker Swarm cluster and a main Traefik load balancer).

I just finished updating DockerSwarm.rocks to use Traefik v2 this weekend, so I updated everything here to also use Traefik v2.

I found a way to test it all even though the official Docker image obviously doesn't have these changes yet.

I deployed a whole cluster of some machines to test it all, debugged (a lot, I'm not a Ruby person :joy:), and fixed the parts that were not working correctly :sweat_smile:, etc.

And finally, I was able to deploy the whole stack, following the same guide, in a couple of commands. Getting automatic HTTPS, and all the stuff described above, etc. :tada: 

---

I also created a test project with a simple test Docker image (just a custom Nginx "Hello World"), and deployed a GitLab Runner to test it all.

And I was able to git clone with the standard port, commit, push.

Then I was able to docker login, and push the docker image.

Also to run CI, building the Docker image in the Runner, and pushing to the registry from the Runner.

And then deploy it from the Runner to the same cluster, pulling from the registry, integrating it all into the same stack with automatic HTTPS, etc.

In the end it all worked great. :sparkles: :rocket: :tada:


tiangolo:
That's awesome! Thanks a lot for all your work here! :rocket: :tada:

solidnerd:
You are welcome! You did also an awesome work.

Issue #2160, opened 2020-05-26
Gitlab backup fail after upgrade PostgreSQL 10 to 11 


nickbug:
After upgrade PostgreSQL 10 to 11, the gitlab backup fail with error  

"**pg_dump version mismatch**"

> ** Execute gitlab:backup:db:create
2020-05-26 17:22:39 +0800 -- Dumping database ...
Dumping PostgreSQL database gitlabhq_production ... pg_dump: server version: 11.8 (Ubuntu 11.8-1.pgdg18.04+1); pg_d                                       ump version: 10.13 (Ubuntu 10.13-1.pgdg18.04+1)
pg_dump: aborting because of server version mismatch
[FAILED]
rake aborted!
Backup::Error: Backup failed

solidnerd:
Hey @nickbug,
it will be available for the short term in `sameersb/gitlab:latest` image and later in the next release. Thanks for reporting it. I'll gonna close this if you Issue still persists please reopen this issue.

PR #2163, opened 2020-05-27
use the latest (12) postgresql-client libraries

sameersbn:
Refer #2160


sameersbn:
IMO it should not hurt to use the postgresql 12 client as it would be backwards compatible with earlier psql server versions

solidnerd:
Also don't see any problems there ;)

PR #2181, opened 2020-06-23
Update gitlab to 13.1.0

Kiina:
https://about.gitlab.com/releases/2020/06/22/gitlab-13-1-released/


Kiina:
Sorry for all the force push spam, but I force pushed a version with the patch removed. Left the code for applying patches as it checkes if the dir exists so in case you ever need it in the future you can still just make the patches directory and put the diff inside. But I can also remove the whole patching logic if preferred.

Had to add the GITALY_WORKHORSE_SOCKET_PATH to keep gitaly from crashing outright after starting the container. Now it's running on my production environment and I haven't seen any issue yet.

f-io:
why did you closed it?

Kiina:
Turns out it's not working. Even after setting the workhorse socket to the right tcp one it doesn't seem to work. Looking through the changelogs and commits I haven't found the reason why

PR #2187, opened 2020-07-02
Upgrade GitLab CE to 13.0.8

f-io:


Kiina:
Could you squash these commits? 12 commits for a version bump seems a bit excessive.

It's probably the best to do 6 commits, one for each component like solidnerd normally does with the commit messages indicating which part got updated

PR #2253, opened 2020-11-08
Reduce container image size

bjf893:
The Gitlab container image is massive. This PR removes unused caches and node_modules from the final image.

The 13.5.3 image on quay.io is 1.6GB, with these changes it should be ~950MB.
 
I have been running this change on my instance for 2 months and have not noticed any issues.


cwildfoerster:
Tested and everything works, reduced from 4.4 to 2.3 GB on my system. LGTM @solidnerd

PR #2261, opened 2020-12-08
Upgrade to 13.6.2

sachilles:
Update to gitlab 13.6.2 (see also https://about.gitlab.com/releases/2020/12/07/security-release-gitlab-13-6-2-released/ )


sachilles:
Unfortunately, I added two additional ampersands. I'm sorry.

cwildfoerster:
Hi @sachilles i think its unnecessary to add the remove the package list, because there is an update and install afterwards and the lists are finally removed at https://github.com/sameersbn/docker-gitlab/blob/a186b0cc01994f38f37a4f57121575dabab0faec/Dockerfile#L57

sachilles:
Hi @cwildfoerster I don't think so, because Docker uses filesystem layers which are stored per command (i.e. RUN, COPY, ...). So it makes sense to keep each filesystem layer "clean".

cwildfoerster:
Okay, thank you, didn't knew that this would blow the final image size. Keep up the good work @sachilles

PR #2279, opened 2020-12-30
Add info about PostgreSQL upgrade

tomastzn:
This change adds more information about upgrading PostgreSQL used with this image.


solidnerd:
LGTM !

PR #2321, opened 2021-03-12
Add LDAP_PREVENT_LDAP_SIGN_IN configuration parameter

l00v3:
Committer: l00v3 <https://github.com/l00v3>
Added configuration parameter LDAP_PREVENT_LDAP_SIGN_IN for "prevent_ldap_sign_in".

Honestly this is my first pull request, I hope I did it the right way. If not, please correct me, so I will know the next time : )

Thank you!


l00v3:
May I kindly ask, when can we expect this to be merged?
Thank you!

f-io:
You can use LDAP_USER_FILTER as an alternativ. I also think it is the better approach to control the access-rights from the directory site not the other way round. You can use for example a filter like (&(objectClass=posixAccount)(memberof=cn=Gitlab,ou=Groups,dc=domain,dc=com)) ...

l00v3:
> You can use LDAP_USER_FILTER as an alternativ. I also think it is the better approach to control the access-rights from the directory site not the other way round. You can use for example a filter like (&(objectClass=posixAccount)(memberof=cn=Gitlab,ou=Groups,dc=domain,dc=com)) ...

Thank you for the reply. Filter is already used for allowing specific accounts to have login permission to this specific service. LDAP_PREVENT_LDAP_SIGN_IN would disable login form, so login would only be allowed via OpenID button.

sameersbn (approved):

sameersbn:
sorry that this was lying here not being merged.

PR #2322, opened 2021-03-12
Upgrade Redis in README to redis:6.2

beli68:
Since 13.0.0, docker-compose.yml uses `redis:5.0.9` for Redis. 
Therefore,  I think it is better to match the README version as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,8 +179,8 @@ Step 2. Launch a redis container
 
 ```bash
 docker run --name gitlab-redis -d \
-    --volume /srv/docker/gitlab/redis:/var/lib/redis \
-    redis:5.0.9
+    --volume /srv/docker/gitlab/redis:/data \
+    redis:6.2
 ```
 
 Step 3. Launch the gitlab container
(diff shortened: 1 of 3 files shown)


beli68:
@sameersbn 
Do I need to do any additional work to merge this PR?

sachilles:
@sameersbn Is there anything that speaks against the incorporation of the adjustments to the README.md file?

sameersbn:
@sachilles if the image change is compatible, then it's good with me

README.md
```diff
@@ -180,7 +180,7 @@ Step 2. Launch a redis container
 ```bash
 docker run --name gitlab-redis -d \
     --volume /srv/docker/gitlab/redis:/var/lib/redis \
```
sameersbn, line 182: according to https://hub.docker.com/_/redis the volume should be mounted at `/data`, so this line would need to be updated

sameersbn (changes requested):

README.md
danyill: Would the volume in the `docker-compose.yml` also need to be modified:

https://github.com/sameersbn/docker-gitlab/blob/09b8c1c265785fcf2e872bda6fcb6709f55b3e4d/docker-compose.yml#L3-L10

```
```

sameersbn: yea, i believe so.
there's also a reference in the Makefile

PR #2331, opened 2021-04-01
v13.9.5

ErwinJunge:
This upgrades to 13.9.5. I don't know if you want to keep the 13.9.x cycle going with tagged builds now that the 13.10.x cycle is started, but just in case: here's an MR :)


ErwinJunge:
FYI: 13.9.4 is unbuildable due to an issue with the mimemagic gem. 13.9.5 fixes that (and a bunch of quite important security issues).

sameersbn (approved):

PR #2330, opened 2021-03-31
[GitLab Security Release] Upgrade to 13.10.1

sachilles:
GitLab Patch Release: 13.10.1 (see <https://about.gitlab.com/releases/2021/03/31/security-release-gitlab-13-10-1-released/>)

This PR is based on #2327 .

PR #2327, opened 2021-03-22
Upgrade to 13.10.0

bjf893:
https://about.gitlab.com/releases/2021/03/22/gitlab-13-10-released/

Gitlab-workhorse is now bundled with gitlab.

PR #2332, opened 2021-04-02
Upgrade to 13.10.2

sachilles:
GitLab Patch Release: 13.10.2 (see <https://about.gitlab.com/releases/2021/04/01/gitlab-13-10-2-released/>)

This PR is based on #2327 and #2330.


bjf893:
@sachilles 
Thanks! Both this PR and #2330 don't work because puma throws an error. 

{"timestamp":"2021-04-02T20:30:47.478Z","pid":2398,"message":"! Unable to load application: LoadError: libmagic.so.1: cannot open shared object file: No such file or directory - /home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/ruby-magic-static-0.3.4/lib/magic/magic.so"}
/home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/ruby-magic-static-0.3.4/lib/magic.rb:3:in `require_relative': libmagic.so.1: cannot open shared object file: No such file or directory - /home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/ruby-magic-static-0.3.4/lib/magic/magic.so (LoadError)

#2331 fixed this for 13.9.5, we need to add libmagic1 as a new dependency

sachilles:
@bjf893 Thanks for pointing to this issue. I'll add libmagic1 as a new dependency to #2330 and rebase this PR.

Issue #2337, opened 2021-04-07
gitlab does not send email - gitlab continer unhealthy status


mradeschnig:
I have the docker-gitlab image 13.10.2 installed on Ubuntu 18.04 (Docker Engine - Community 20.10.5).
First I have the problem that I could not test sending email which I solved with another incident. Sending
emails from the gitlab console works. 
But still I do not get emails for passwort recovery as well no welcome mail after registration is sent to the new user.

How to find out why the gitlab container is in unhealthy state and second whats the problem
that gitlab does not send emails.

Obviously I don't see anything wrong except maybe this line in the logs ( I use portainer so I guess this aggregates all logs together):
/sbin/entrypoint.sh: line 19:   666 Terminated              /usr/bin/supervisord -nc /etc/supervisor/supervisord.conf

Kind Regards

kkimurak:
Sorry for not helpful about email, but current container always reports unhealthy if you're running with relative URL (I'm about to submit a PR later).  
And, I also sometimes see such logs even everything works fine.

mradeschnig:
I use a relative url (/git ) than this might be the reason for the heathstatus = unhealthy.
Thank you.

mradeschnig:
A friend helped me. He think i made it wrong to use the internal server name as GITLAB_HOST instead the external  url.
And in fact this was right. Email works now, but all the actions to pull/clone the repositories also have now the wrong url
and can not be reached from external. I use an Apache ReverseProxy to redirect the external call to the correct server.
So in fact what is the correct proceeding? Or do I need to change something for the ReverseProxy? In fact it is strange as the mail server is external.

mradeschnig:
The error (misconfiguration) is solved. It was the cofig: I set the GITLAB_HOST= https://subdomain.mydomain.com
but it only works with GITLAB_HOST= subdomain.mydomain.com.
I realized it when I put the local hostname there.
I wonder that it is not checked during creation or protokol is cut off.
Some functions work, some like email won't work with the protokol included.
Maybe a check would make sense the it is a correct hostname.

PR #2338, opened 2021-04-08
Support healthcheck with relative URL

kkimurak:
partially solves #2337 .

Current healthcheck script (which run `curl http://localhost/-/liveness`) always reports unhealthy in the case relative URL in use.  
This script is [generated on build time](https://github.com/sameersbn/docker-gitlab/blob/82630d2d93e424467c1ab9d5f43918f1a8ba023a/assets/build/install.sh#L426-L435), and never updated. So if the user set `GITLAB_RELATIVE_URL_ROOT`, URL mismatch will happen and returns 404.

This PR transplants script generation to runtime to allow us to use `${GITLAB_RELATIVE_URL_ROOT}` variable to set healthcheck URL .
Nothing will happen for those who don't use relative URL since the variable will be replaced by empty string.

example output (`GITLAB_RELATIVE_URL: /gitlab` in docker-compose.yml)

```
$ docker exec -it gitlab /bin/bash
git@0468dd640abe: /home/git/gitlab# cat /usr/local/sbin/healthcheck
#!/bin/bash
url=http://localhost/gitlab-/liveness
options=( '--insecure' '--location' '--silent' )
curl "${options[@]}" $url
[[ "$(curl ${options[@]} -o /dev/null -I -w '%{http_code}' $url)" == "200" ]]
```

This is first time I submit a PR for this repository, so please let me know if I'm doing something weird.

### DIFF
--- a/assets/build/install.sh
+++ b/assets/build/install.sh
@@ -423,17 +423,6 @@ programs=sshd,nginx,mail_room,cron
 priority=20
 EOF
 
-# configure healthcheck script
-## https://docs.gitlab.com/ee/user/admin_area/monitoring/health_check.html
-cat > /usr/local/sbin/healthcheck <<EOF
-#!/bin/bash
-url=http://localhost/-/liveness
-options=( '--insecure' '--location' '--silent' )
-curl "\${options[@]}" \$url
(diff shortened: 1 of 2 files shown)


sameersbn (approved):
lgtm

Issue #2375, opened 2021-06-28
Pages service does not start after upgrading from 13.12.4 to 14.0.0


Aoiujz:
Pages service does not start after upgrading from 13.12.4 to 14.0.0，then upgraded to 14.0.1 and the same problem occurred.

```
time="2021-06-28T01:01:42Z" level=info msg="GitLab Pages Daemon" revision=0e711de version=1.40.0
time="2021-06-28T01:01:42Z" level=info msg="URL: https://gitlab.com/gitlab-org/gitlab-pages"
time="2021-06-28T01:01:42Z" level=fatal msg="could not create domains config source" error="GitLab API URL or API secret has not been provided"
time="2021-06-28T01:01:43Z" level=info msg="GitLab Pages Daemon" revision=0e711de version=1.40.0
time="2021-06-28T01:01:43Z" level=info msg="URL: https://gitlab.com/gitlab-org/gitlab-pages"
time="2021-06-28T01:01:43Z" level=fatal msg="could not create domains config source" error="GitLab API URL or API secret has not been provided"
time="2021-06-28T01:01:45Z" level=info msg="GitLab Pages Daemon" revision=0e711de version=1.40.0
time="2021-06-28T01:01:45Z" level=info msg="URL: https://gitlab.com/gitlab-org/gitlab-pages"
time="2021-06-28T01:01:45Z" level=fatal msg="could not create domains config source" error="GitLab API URL or API secret has not been provided"
time="2021-06-28T01:01:48Z" level=info msg="GitLab Pages Daemon" revision=0e711de version=1.40.0
time="2021-06-28T01:01:48Z" level=info msg="URL: https://gitlab.com/gitlab-org/gitlab-pages"
time="2021-06-28T01:01:48Z" level=fatal msg="could not create domains config source" error="GitLab API URL or API secret has not been provided"
time="2021-06-28T02:28:27Z" level=info msg="GitLab Pages Daemon" revision=0e711de version=1.40.0
time="2021-06-28T02:28:27Z" level=info msg="URL: https://gitlab.com/gitlab-org/gitlab-pages"
time="2021-06-28T02:28:27Z" level=fatal msg="could not create domains config source" error="GitLab API URL or API secret has not been provided"
time="2021-06-28T02:28:28Z" level=info msg="GitLab Pages Daemon" revision=0e711de version=1.40.0
[...]

PR #2389, opened 2021-07-20
Resolved. GitLab Pages doesn't work since GitLab v14.0

sue445:
ref. #2375

As I can see from the error message below, it looks like both
`api-secret-key` and `gitlab-server` (or `internal-gitlab-server` ) are required since GitLab v14.0

```
time="2021-06-28T02:28:34Z" level=fatal msg="could not create domains config source" error="GitLab API URL or API secret has not been provided"
```

https://gitlab.com/gitlab-org/gitlab-pages/-/blob/v1.41.0/internal/source/gitlab/client/client.go#L50-52

This patch is actually working on my GitLab


assets/runtime/config/gitlab-pages/config
```diff
@@ -4,3 +4,5 @@ auth-redirect-uri={{GITLAB_PAGES_ACCESS_REDIRECT_URI}}
 auth-secret={{GITLAB_PAGES_ACCESS_SECRET}}
 gitlab-server={{GITLAB_PAGES_ACCESS_CONTROL_SERVER}}
 artifacts-server={{GITLAB_PAGES_ARTIFACTS_SERVER_URL}}
+internal-gitlab-server=http://localhost:8181
+api-secret-key=/home/git/gitlab/.gitlab_pages_secret
```
achretien, line 8: you should use GITLAB_INSTALL_DIR here too match config in assets/runtime/functions

sue445: @achretien Fixed at 73dae28

icedream:
@sameersbn I saw that you updated to 14.0.6 just recently. Can you take a look at this and possibly include this in the next update at least? This affects a production environment I'm maintaining right now and it seems to be affecting all users of the GitLab Pages feature.

sameersbn (approved):

danyill:
When I manually patch these changes into the v14.1.0 container pages works fine. Thanks for this :+1:

PR #2393, opened 2021-07-25
Fix #2392 - update comments about "_gitlab-data" volume mount point.

### LINKED ISSUE #2392: SSL topic on README.md  is outdated because the way docker-compose volumes are defined (Not binded anymore)
the README file is outdated about volumes

> GitLab is a code hosting software and as such you don't want to lose your code when the docker container is stopped/deleted. To avoid losing any data, you should mount a volume at,
>
> /home/git/data
>
> Note that if you are using the docker-compose approach, this has already been done for you.

The last line is not true since f37514ecf8c8cf42856786b2d3fef00e1d046a0a

And the topic about SSL  at 

> If you remember from above, the /home/git/data path is the path of the data store, which means that we have to create a folder named certs/ inside /srv/docker/gitlab/gitlab/ and copy the files into it and as a measure of security we'll update the permission on the gitlab.key file to only be readable by the owner.
>
> ``` bash
>mkdir -p /srv/docker/gitlab/gitlab/certs
>cp gitlab.key /srv/docker/gitlab/gitlab/certs/
>cp gitlab.crt /srv/docker/gitlab/gitlab/certs/
>cp dhparam.pem /srv/docker/gitlab/gitlab/certs/
>chmod 400 /srv/docker/gitlab/gitlab/certs/gitlab.key
```
I believe the paths are incorrect 

it should be something like 

(note the "root_" prefix on volume's path and name, if you are running docker as root)

```bash
$> docker volume inspect root_gitlab-data
```
result looks like  ... check the path where it has been mounted

```json
[
    {
        "CreatedAt": "2021-07-25T02:47:41Z",
        "Driver": "local",
        "Labels": {
            "com.docker.compose.project": "root",
            "com.docker.compose.version": "1.25.0",
            "com.docker.compose.volume": "gitlab-data"
        },
        "Mountpoint": "/var/lib/docker/volumes/root_gitlab-data/_data",
        "Name": "root_gitlab-data",
        "Options": null,
        "Scope": "local"
    }
]
```

and then copy the `certs` directory inside the volumes (Moountpoint) path

cadu-leite:


sameersbn (approved):

sameersbn:
thx for the contribution

cadu-leite:
glad to help

PR #2404, opened 2021-08-24
Upgrade Gitlab CE to 14.2.0

sachilles:
GitLab Patch Release: 14.2.0 (see <https://about.gitlab.com/releases/2021/08/22/gitlab-14-2-released/>)

**Hint:** An adjustment in the file 'assets/build/patches/0001-fix-feature-checking-for-gitaly-on-a-fresh-install.patch' was necessary, otherwise the patch was not applicable and therefore the build process was not successful.

Do we still need the fix? There are no such adjustments in the omnibus image repo.


sameersbn:
@sachilles looks like the patches are getting applied cleanly

sachilles:
Hi,

my question is whether the patch is still necessary at all.

sameersbn:
> my question is whether the patch is still necessary at all.

Sorry, I am really not sure why it was added and if it could be removed.

closed by #2406

PR #2423, opened 2021-10-04
Update Gitlab to 14.2.5

avvertix:
- gitlab: upgrade CE to v14.2.5
- gitaly: upgrade to v14.2.5
- gitlab-pages: upgrade to v1.45.0

https://about.gitlab.com/releases/2021/09/30/security-release-gitlab-14-3-1-released/


omueller:
thanks for the update @avvertix !  It would be great if we could get an "official" release, as the last one was nearly 2 months ago.
(GitLab 14.1.1-1, 20210803), do you think it would be possible sometime later this month @sameersbn & team ?  I am a bit worried to try an "unofficial"  (from the newer tags) after browsing the latest issues (db/migration issues).   

Thanks/merci & a nice week to you!

avvertix:
@omueller I'm not sure how @sameersbn plan the releases, but I'm using the tag 14.2.4 since a week and seems stable enough to be used. Of course this is just my opinion and experience

omueller:
@avvertix ok, that is already great to know, thanks!   I will give it a try next week-end then.

sachilles:
@avvertix I noticed that you are using Gitlab Pages version 1.45.0. According to <https://gitlab.com/gitlab-org/gitlab-foss/-/blob/v14.3.1/GITLAB_PAGES_VERSION> the recommended version is 1.42.0.

What is the reason for choosing a different version?

avvertix:
I think I checked erroneously the wrong branch. Will revert to 1.42.0

PR #2427, opened 2021-10-09
suppress build time warnings

kkimurak:
This patch suppresses warnings during `docker build`.

- [git] set `advice.detachHead=false` :  
  GitLab releases are based on tag, is detached head, not branch. `git` warns so when doing so.
- [bundle] use `bundle config set --local` instead of those [deprecated arguments](https://github.com/rubygems/bundler/blob/master/UPGRADING.md#cli-deprecations) for `bundle install` :  
  - `--deployment`
  - `--with` / `--without`


sachilles:
Hi,

I'll merge your PR later in the evening. Sorry for the delay.

PR #2425, opened 2021-10-08
Update Gitlab to 14.3.2

tvilleminot:
**14.3.2**
- gitlab: upgrade CE to v14.3.2
- gitaly: upgrade to v14.3.2
- gitlab-shell: upgrade to v13.20.0
- gitlab-pages: upgrade to v1.44.0
- ubuntu: upgrade to focal-20210921

https://about.gitlab.com/releases/2021/10/01/gitlab-14-3-2-released/

I am using version 13.20.0 of gitlab-shell because version 13.21.1 causes the following error:

`go: inconsistent vendoring in /home/git/gitlab-shell:
	github.com/grpc-ecosystem/go-grpc-middleware@v1.3.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/grpc-ecosystem/go-grpc-prometheus@v1.2.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/mattn/go-shellwords@v1.0.11: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/mikesmitty/edkey@v0.0.0-20170222072505-3356ea4e686a: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/otiai10/copy@v1.4.2: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/pires/go-proxyproto@v0.6.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/prometheus/client_golang@v1.10.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/stretchr/testify@v1.7.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	gitlab.com/gitlab-org/gitaly/v14@v14.0.0-rc1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	gitlab.com/gitlab-org/labkit@v1.7.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/crypto@v0.0.0-20201221181555-eec23a3978ad: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
[...]

PR #2435, opened 2021-10-19
upgrade GitLab CE to 14.3.0

sachilles:
GitLab Release: 14.3.0 (see https://about.gitlab.com/releases/2021/09/22/gitlab-14-3-released/)

### DIFF
--- a/assets/build/install.sh
+++ b/assets/build/install.sh
@@ -111,7 +111,7 @@ exec_as_git bundle config set --local deployment 'true'
 exec_as_git bundle config set --local with 'development test'
 exec_as_git bundle install -j"$(nproc)"
 exec_as_git "${GOROOT}"/bin/go mod vendor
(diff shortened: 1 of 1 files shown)


kkimurak:
Also, I remembered that build has fail during compiling `gitlab-shell`. We need to add trick for `install.sh` as I suggested in https://github.com/sameersbn/docker-gitlab/pull/2425#discussion_r728476081

sachilles:
Oh my god, this is getting embarrassing. If I hadn't already tested this separately... Now let's keep fingers crossed.

sachilles:
If the build works, I will create a PR for Gitlab 14.3.1 right afterwards. The corresponding branch is already prepared.

sachilles:
@sameersbn Could you be so kind to publish the Docker images so that the release can be published?

PR #2477, opened 2021-12-24
Upgrade GitLab CE to 14.6.0

sachilles:
GitLab Release: 14.6.0 (see <https://about.gitlab.com/releases/2021/12/22/gitlab-14-6-released/>)


sachilles:
Because of the deprecation warning during startup of the gitlab docker image,

```
WARNING: This installation of GitLab uses a deprecated syntax for 'config/database.yml'. The support for this syntax will be removed in 15.0. More information can be found here: https://gitlab.com/gitlab-org/gitlab/-/issues/338182
```
, the keyword main was added to the configuration file for the database (`config/database.yml`).

For more details see <https://gitlab.com/gitlab-org/gitlab/-/issues/338182> and <https://docs.gitlab.com/ee/development/database/multiple_databases.html#configdatabaseyml>.

Issue #2482, opened 2022-01-02
Assigning issues after upgrading to >= 14.4 raise an error (but it gets assigned)


paolomainardi:
As reported here: https://gitlab.com/gitlab-org/gitlab/-/issues/349481

Anyone is having the same issue ?

paolomainardi:
What i've noticed now is that opening issues raise this error:

```
Finished "/-/cable/" [WebSocket] for 5.90.96.187 at 2022-01-02 23:09:26 +0100
/home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/actioncable-6.1.4.1/lib/action_cable/server/configuration.rb:27:in `pubsub_adapter'
/home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/actioncable-6.1.4.1/lib/action_cable/server/base.rb:83:in `block in pubsub'
/home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/actioncable-6.1.4.1/lib/action_cable/server/base.rb:83:in `synchronize'
/home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/actioncable-6.1.4.1/lib/action_cable/server/base.rb:83:in `pubsub'
/home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/actioncable-6.1.4.1/lib/action_cable/connection/base.rb:53:in `pubsub'
/home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/actioncable-6.1.4.1/lib/action_cable/channel/streams.rb:140:in `pubsub'
/home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/actioncable-6.1.4.1/lib/action_cable/channel/streams.rb:122:in `block in stop_all_streams'
/home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/actioncable-6.1.4.1/lib/action_cable/channel/streams.rb:121:in `each'
/home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/actioncable-6.1.4.1/lib/action_cable/channel/streams.rb:121:in `stop_all_streams'
/home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/actioncable-6.1.4.1/lib/action_cable/engine.rb:68:in `block (3 levels) in <class:Engine>'
[...]

PR #2483, opened 2022-01-03
Configure ActionCable

paolomainardi:
Related to: https://github.com/sameersbn/docker-gitlab/issues/2482

### DIFF
--- a/assets/runtime/config/gitlabhq/cable.yml
+++ b/assets/runtime/config/gitlabhq/cable.yml
@@ -1,0 +1,14 @@
+# This is a template taken from here:
+# https://gitlab.com/gitlab-org/gitlab/-/blob/master/config/cable.yml.example
+development:
+  adapter: redis
+  url: redis://localhost:6379
+  channel_prefix: gitlab_development
+test:
+  adapter: redis
+  url: redis://localhost:6379
+  channel_prefix: gitlab_test
+production:
+  adapter: redis
+  url: redis://{{REDIS_HOST}}:{{REDIS_PORT}}/{{REDIS_DB_NUMBER}}
+  channel_prefix: gitlab_production
--- a/assets/runtime/functions
+++ b/assets/runtime/functions
@@ -19,6 +19,7 @@ GITLAB_PUMA_CONFIG="${GITLAB_INSTALL_DIR}/config/puma.rb"
 GITLAB_RELATIVE_URL_CONFIG="${GITLAB_INSTALL_DIR}/config/initializers/relative_url.rb"
 GITLAB_SMTP_CONFIG="${GITLAB_INSTALL_DIR}/config/initializers/smtp_settings.rb"
 GITLAB_RESQUE_CONFIG="${GITLAB_INSTALL_DIR}/config/resque.yml"
+GITLAB_ACTIONCABLE_CONFIG="${GITLAB_INSTALL_DIR}/config/cable.yml"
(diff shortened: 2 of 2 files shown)


assets/runtime/functions
```diff
@@ -232,6 +233,17 @@ gitlab_configure_redis() {
...
 
+gitlab_configure_actioncable() {
+  echo -n "Configuring gitlab::actioncable"
+
+  gitlab_finalize_redis_parameters
+  gitlab_check_redis_connection
+
+  update_template ${GITLAB_ACTIONCABLE_CONFIG} \
+    REDIS_HOST \
+    REDIS_PORT
```
achretien, line 244: I think you need to add REDIS_DB_NUMBER here

paolomainardi: nice catch, added

assets/runtime/config/gitlabhq/cable.yml
```diff
@@ -0,0 +1,14 @@
+# This is a template taken from here:
+# https://gitlab.com/gitlab-org/gitlab/-/blob/master/config/cable.yml.example
+development:
```
sachilles, line 3: If the definition of the Redis connection for the environments 'development' and 'test' is not needed, then I would suggest to either use other "default values" (like
`url: redis://:fakepass@fake.redis.actioncable.com:1234` and `channel_prefix: gitlab_fake_development`) or not be defined at all.

assets/runtime/functions
sachilles: Nice work!

assets/runtime/config/gitlabhq/cable.yml
```diff
@@ -0,0 +1,14 @@
...
+  adapter: redis
+  url: redis://localhost:6379
+  channel_prefix: gitlab_development
+test:
+  adapter: redis
+  url: redis://localhost:6379
+  channel_prefix: gitlab_test
```
sachilles, line 10: If the definition of the Redis connection for the environments 'development' and 'test' is not needed, then I would suggest to either use other "default values" (like
`url: redis://:fakepass@fake.redis.actioncable.com:1234` and `channel_prefix: gitlab_fake_development`) or these environmental setups should not be defined at all.

paolomainardi: Yes at the beginning I haven’t defined them, but then I saw this file: https://github.com/sameersbn/docker-gitlab/blob/09e76bb2e7d4c805bfc2cb70cb1b0ca89cd60216/assets/runtime/config/gitlabhq/resque.yml and replicated the same behavior. Anyway it is quite harmless to keep the other two, it’s just for consistency with other files.

sachilles: Okay, that sounds reasonable. Have you already tested this version and can you confirm that the error in the `production.log` file no longer occurs?

sachilles (approved):

assets/runtime/config/gitlabhq/cable.yml
paolomainardi: @sachilles yes sure, using it in production now, it works like a charm.

sachilles: That sounds good. Thanks for your contribution! A patch release will follow soon.

paolomainardi: The bug affects also 14.4 and 14.5 @sachilles 

thanks for merging!

PR #2488, opened 2022-01-13
Upgrade GitLab CE to 14.6.2

sachilles:
GitLab Security Release: 14.6.2 (see <https://about.gitlab.com/releases/2022/01/11/security-release-gitlab-14-6-2-released/>)


Dockerfile
```diff
@@ -1,14 +1,14 @@
 FROM ubuntu:focal-20211006
```
kkimurak, line 1: Newer version `focal-20220105` is available.  
https://hub.docker.com/layers/ubuntu/library/ubuntu/focal-20220105/images/sha256-64162ac111b666daf1305de1888eb67a3033f62000f5ff781fe529aff8f88b09?context=explore

kkimurak: Sounds good. 

We can use [`--pull` option for `docker build`](https://docs.docker.com/engine/reference/commandline/build/#options) command. With this option, docker always attempt to pull a newer version of the image. So we can use `ubuntu:focal` instead of specific date based tag without worrying about cache.  

In this repository, `docker build` commands are used in [`Makefile`](https://github.com/sameersbn/docker-gitlab/blob/ca3338629df24b798f82ce9cac23677b6eaff632/Makefile##L13-L16) and [`.circleci/config.yml`](https://github.com/sameersbn/docker-gitlab/blob/ca3338629df24b798f82ce9cac23677b6eaff632/.circleci/config.yml#L209) so they should be updated as well.  
Personally I think such change should have individual PR (It would be really strange that build configurations would be changed by just updating target version). But collaborator is you, so I'll leave the final decision.

achretien: if you use latest ubuntu image and rebuild a tag in a few month you will have a different version of the base image.

Anyway it is not the point of this PR :)

Issue #2491, opened 2022-01-16
"EMFILE: too many open files" error at gitlab:assets:compile stage


stephane-klein:
When I execute:

```
$ make release
```
On MacOS ou Fedora 35, I have this error:

```
$ NODE_OPTIONS="--max-old-space-size=3584" webpack --config config/webpack.config.js
IncrementalWebpackCompiler: Status – disabled
Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db

Why you should do it regularly:
https://github.com/browserslist/browserslist#browsers-data-updating
Warning: fragment with name PageInfo already exists.
graphql-tag enforces all fragment names across your application to be unique; read more about
this in the docs: http://dev.apollodata.com/core/fragments.html#unique-names
Error: EMFILE: too many open files, open '/home/git/gitlab/node_modules/.cache/compression-webpack-plugin/tmp/64ab441a'
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Error: Unable to compile webpack production bundle.
The command '/bin/sh -c bash ${GITLAB_BUILD_DIR}/install.sh' returned a non-zero code: 1
make: *** [Makefile:14 : build] Erreur 1
```

I managed to correct this error with this patch: add `NO_SOURCEMAPS=true` [here](https://github.com/sameersbn/docker-gitlab/blob/f357effe3dade1483ee22258598003a65d482f9e/assets/build/install.sh#L187):

```
echo "Compiling assets. Please be patient, this could take a while..."
exec_as_git bundle exec rake gitlab:assets:compile USE_DB=false SKIP_STORAGE_VALIDATION=true NO_SOURCEMAPS=true NODE_OPTIONS="--max-old-space-size=4096"
```

I am writing a Pull Request.

sachilles:
Hi,

I'm not sure if we should change the way we compile the Ruby assets, as I checked the logfiles on different platforms (Debian, Ubuntu, Archlinux, including our Circle Ci) and did not find the bug you reported.

However, searching a bit on the Internet, I found some references to the bug you reported related to file limits:

* node and Error: EMFILE, too many open files (<https://stackoverflow.com/questions/8965606/node-and-error-emfile-too-many-open-files>)
* OS X: "Error: EMFILE: too many open files, watch" when running tests (<https://github.com/facebook/create-react-app/issues/4540>)
[...]

PR #2492, opened 2022-01-16
Disable sourcemaps generation to fix "EMFILE: too many open files" er…Disable sourcemaps generation to fix "EMFILE: too many open files" error #2491

stephane-klein:
Hello @sachilles, 

here is a patch to disable sourcemaps generation to fix "EMFILE: too many open files" error #2491

What do you think?

Best regards,
Stéphane


sachilles:
As in #2491 stated, the problem was solved by using docker instead of moby (under fedora). Thus, this PR will be closed without merging.

PR #2510, opened 2022-02-24
Upgrade GitLab CE to 14.8.1

sachilles:
GitLab Patch Release: 14.8.1 (see <https://about.gitlab.com/releases/2022/02/23/gitlab-14-8-1-released/>)


tDo:
hi @sachilles it seems like the 14.8.1 tag is missing from both github and for the docker-images.
Would it be possible to add it? (Switched over to latest for now, since the failing backups bug in 14.8.0 is a nasty one...)

Thank you in advanced :)

sachilles:
@tDo I fixed it. Thanks for your hint.

PR #2511, opened 2022-02-26
Fix: HealthCheck Script will Fail if the ENV GITLAB_HTTPS set True

YouthLab:
The liveness probe request will return a 302 redirect status code when enable https endpoint,
Which will cause a misleading healthcheck result.
Capture

### DIFF
--- a/assets/runtime/functions
+++ b/assets/runtime/functions
@@ -1641,9 +1641,13 @@ update_ssh_listen_port() {
 generate_healthcheck_script() {
   # configure healthcheck script
   ## https://docs.gitlab.com/ee/user/admin_area/monitoring/health_check.html
+  local HEALTHCHECK_PROTOCOL="http"
+  if [[ "${GITLAB_HTTPS}" == true ]]; then
+    HEALTHCHECK_PROTOCOL="${HEALTHCHECK_PROTOCOL}s"
+  fi
 cat > /usr/local/sbin/healthcheck <<EOF
 #!/bin/bash
-url=http://localhost${GITLAB_RELATIVE_URL_ROOT}/-/liveness
+url=${HEALTHCHECK_PROTOCOL}://localhost${GITLAB_RELATIVE_URL_ROOT}/-/liveness
 options=( '--insecure' '--location' '--silent' )
 curl "\${options[@]}" \$url
 [[ "\$(curl \${options[@]} -o /dev/null -I -w '%{http_code}' \$url)" == "200" ]]


YouthLab:
@sameersbn @solidnerd  Hi～ guys，Please help me to review the PR. ：）

solidnerd:
ping @sachilles ;) . Since he is the most active maintainer.

sachilles:
@YouthLab I'll check your PR later this weekend. ;-)

YouthLab:
@sachilles Hi, Steven, just remind you to check the code. 😃
I suggest you review the other one (PR: Upgrade to 14.8.2)too,so that the fix will release with the latest version.
Thanks!

sachilles:
Thanks for your reminder and understanding. 😄

assets/runtime/functions
```diff
@@ -1643,7 +1643,11 @@ generate_healthcheck_script() {
...
 #!/bin/bash
-url=http://localhost${GITLAB_RELATIVE_URL_ROOT}/-/liveness
+if [[ $GITLAB_HTTPS == true ]]; then
+  url=https://localhost${GITLAB_RELATIVE_URL_ROOT}/-/liveness
+else
+  url=http://localhost${GITLAB_RELATIVE_URL_ROOT}/-/liveness
+fi
 options=( '--insecure' '--location' '--silent' )
 curl "\${options[@]}" \$url
 [[ "\$(curl \${options[@]} -o /dev/null -I -w '%{http_code}' \$url)" == "200" ]]
```
sachilles, line 1653: I am a little puzzled. Shouldn't the `--location` option ensure that the redirect is followed automatically?

The man page of curl says:

```
       -k, --insecure
              (TLS) By default, every SSL connection curl makes is verified to be secure. This option allows curl to proceed and operate even  for  server
              connections otherwise considered insecure.

              The  server  connection is verified by making sure the server's certificate contains the right name and verifies successfully using the cert
              store.

              See this online resource for further details:
               https://curl.haxx.se/docs/sslcerts.html

...

       -L, --location
              (HTTP) If the server reports that the requested page has moved to a different location (indicated with a Location: header and a 3XX response
              code), this option will make curl redo the request on the new place. If used together with -i, --include or -I, --head, headers from all re‐
              quested pages will be shown. When authentication is used, curl only sends its credentials to the initial host. If a redirect takes curl to a
              different host, it won't be able to intercept the user+password. See also --location-trusted on how to change this. You can limit the amount
              of redirects to follow by using the --max-redirs option.

              When curl follows a redirect and the request is not a plain GET (for example POST or PUT), it will do the following request with  a  GET  if
              the  HTTP response was 301, 302, or 303. If the response code was any other 3xx code, curl will re-send the following request using the same
              unmodified method.

              You can tell curl to not change the non-GET request method to GET after a 30x response by using the dedicated options for  that:  --post301,
              --post302 and --post303.

...

       -s, --silent
              Silent or quiet mode. Don't show progress meter or error messages.  Makes Curl mute. It will still output the data you ask for,  potentially
              even to the terminal/stdout unless you redirect it.

              Use -S, --show-error in addition to this option to disable progress meter but still show error messages.

              See also -v, --verbose and --stderr.
```

sachilles (commented):
I am trying to understand the problem. Does anyone have experience with the special constellation with the parameters of `curl`?

assets/runtime/functions
```diff
@@ -1643,7 +1643,11 @@ generate_healthcheck_script() {
   ## https://docs.gitlab.com/ee/user/admin_area/monitoring/health_check.html
 cat > /usr/local/sbin/healthcheck <<EOF
 #!/bin/bash
-url=http://localhost${GITLAB_RELATIVE_URL_ROOT}/-/liveness
+if [[ $GITLAB_HTTPS == true ]]; then
```
sachilles, line 1646: Could you be so kind and use variable expansion by using curly braces, i.e. please replace 

```
if [[ $GITLAB_HTTPS == true ]]; then
```

by

```
if [[ ${GITLAB_HTTPS} == true ]]; then
```

Just to be explicit.

YouthLab: DONE. With pleasure

assets/runtime/functions
YouthLab: I don't known the extra parameters '--location' will handle redirect before.
After you post these info, I did a deep diagram and i found the real problem：

I set the `GITLAB_PORT` ENV to 10443, then the liveness probe will redirect to localhost:10443, but the port have not be listening by local gitlab. Which will cause a ` Failed to connect to localhost port 10443: Connection refused` Err.

I think the solution is either keep the container‘s port be the same with 443 or
use my pr :)   (keep use localhost to be the hostname rather use the public hostname is more effective.)

YouthLab: @sachilles

kkimurak: Generally looks good. Thank you for reporting issue about redirect when `GITLAB_PORT` is set.

I think it is enough to check `GITLAB_HTTPS` before generating the script, not in the script. `GITLAB_HTTPS` is set at container startup and will not change during execution. 

```sh
generate_healthcheck_script() {
  # configure healthcheck script
  ## https://docs.gitlab.com/ee/user/admin_area/monitoring/health_check.html
  local HEALTHCHECK_PROTOCOL="http"
  if [[ "${GITLAB_HTTPS}" == true ]]; then
    HEALTHCHECK_PROTOCOL="${HEALTHCHECK_PROTOCOL}s"
  fi
cat > /usr/local/sbin/healthcheck <<EOF
#!/bin/bash
url=${HEALTHCHECK_PROTOCOL}://localhost${GITLAB_RELATIVE_URL_ROOT}/-/liveness
options=( '--insecure' '--location' '--silent' )
curl "\${options[@]}" \$url
[[ "\$(curl \${options[@]} -o /dev/null -I -w '%{http_code}' \$url)" == "200" ]]
EOF
  chmod +x /usr/local/sbin/healthcheck
}
```

sachilles: Hi, sorry for the very late reply. kkimurak's suggestion has charm. Could you adjust your PR so that it can be merged in the near future?

sachilles:
@YouthLab please check and realize the suggestion by kkimurak.

sachilles:
@YouthLab Please replace your suggested change to

```
generate_healthcheck_script() {
  # configure healthcheck script
  ## https://docs.gitlab.com/ee/user/admin_area/monitoring/health_check.html
  local HEALTHCHECK_PROTOCOL="http"
  if [[ "${GITLAB_HTTPS}" == true ]]; then
    HEALTHCHECK_PROTOCOL="${HEALTHCHECK_PROTOCOL}s"
  fi
cat > /usr/local/sbin/healthcheck <<EOF
#!/bin/bash
url=${HEALTHCHECK_PROTOCOL}://localhost${GITLAB_RELATIVE_URL_ROOT}/-/liveness
options=( '--insecure' '--location' '--silent' )
curl "\${options[@]}" \$url
[[ "\$(curl \${options[@]} -o /dev/null -I -w '%{http_code}' \$url)" == "200" ]]
EOF
  chmod +x /usr/local/sbin/healthcheck
}
```

as suggested by @kkimurak as soon as possible. Only then can I merge your PR.

YouthLab:
Already replaced.

sachilles:
Thanks. LGTM.

PR #2516, opened 2022-03-06
support GITLAB_ISSUE_CLOSING_PATTERN

### LINKED ISSUE #2408: Add issue_closing_pattern parameters for ENV
Hi!
Add please https://docs.gitlab.com/ee/administration/issue_closing_pattern.html support for parameters in ENV.
If i edit gitlab.yml it has no effect  because will be updated and overwritten by entrypoint.sh on restart.

kkimurak:
close #2408 
This PR make `gitlab`->`issue_closing_pattern` accessible from environment variable `GITLAB_ISSUE_CLOSING_PATTERN`.


kkimurak:
NOTE: Default value is not rendered correctly in README.md because it contains `|`, is recognized as table delimiter. This can be corrected by escaping all of `|` like `\|`. But I'm not sure it is okay because it may confusing someone who reads document locally, without rendering. 

raw markdown for GITLAB_ISSUE_CLOSING_PATTERN description that contains "|" in cell

a regex string containing `|` does not correctly rendered in github web view

I have proposal to expand the table into a format with parameter names in the header and description in plain text. See the [branch](https://github.com/kkimurak/docker-gitlab/tree/reorganize-configuration-parameters-in-readme) in my repo.

kkimurak:
Hmm, something going bad during test. it says that nginx returns 502 bad gateway. I'll check

kkimurak:
It seem to must not to be enclosed in double quotes. I replaced it with single quotes and the test passed.

sachilles:
Before we explode the whole table, wouldn't it be more advisable to define the regular expressions outside the table and refer to them in the table, isn't it?

README.md
```diff
@@ -807,6 +807,7 @@ Below is the complete list of available options that can be used to customize yo
 | `GITLAB_EMAIL_SMIME_KEY_FILE` | Specifies the path to a S/MIME private key file in PEM format, unencrypted. Defaults to ``. |
 | `GITLAB_EMAIL_SMIME_CERT_FILE` | Specifies the path to a S/MIME public certificate key in PEM format. Defaults to ``. |
 | `GITLAB_DEFAULT_THEME` | Default theme ID, by default 2. (1 - Indigo, 2 - Dark, 3 - Light, 4 - Blue, 5 - Green, 6 - Light Indigo, 7 - Light Blue, 8 - Light Green, 9 - Red, 10 - Light Red) |
+| `GITLAB_ISSUE_CLOSING_PATTERN` | Issue closing pattern regex. See [GitLab's documentation](https://docs.gitlab.com/ee/administration/issue_closing_pattern.html) for more detail. Defaults to ` \b((?:[Cc]los(?:e[sd]?|ing)|\b[Ff]ix(?:e[sd]|ing)?|\b[Rr]esolv(?:e[sd]?|ing)|\b[Ii]mplement(?:s|ed|ing)?)(:?) +(?:(?:issues? +)?%{issue_ref}(?:(?:, *| +and +)?)|([A-Z][A-Z0-9_]+-\d+))+) ` . |
```
sachilles, line 810: I propose to mention the default values outside this table e.g. directly after this table as an additional hint and with a reference.

I know, in any case our Readme.md has to be cleaned up and restructured. I guess this will need some time. 
```suggestion
| `GITLAB_ISSUE_CLOSING_PATTERN` | Issue closing pattern regex. See [GitLab's documentation](https://docs.gitlab.com/ee/administration/issue_closing_pattern.html) for more detail. The default values can be found directly after this table below [LINK]. |
```

What do you think?

sachilles: An example of well structured Readme.md is given by <https://github.com/spantaleev/matrix-docker-ansible-deploy>.

However, this does not make finding information any easier ^^.

sachilles: I tried a few more tricks, however, resolving the table is the most efficient (and also readable) option. Could you put the individual table entries into a list in the form

- Important parameter 1`: Important explanation of parameter 1.
- Important parameter 2': Important explanation of parameter 2.

formatted?

sachilles: I would like to mix this PR together with #2531. I hope I'm not pushing you with this. (However, do you need my assistance?)

kkimurak: Hi. sorry for late.  

Another reason I chose that format is that most of markdown processors generate anchor to the header. Currently when I explain individual parameters, I sometimes find it inconvenient because there is no suitable link other than ″configuration parameters″ even though there are so many parameters.

But list style you suggested is also looks smart. Im wondering which is better. Could you decide which one looks better, considering the reasons given above? I respect your judgment and submit a pr.

sachilles:
Hello, the automatically generated anchor is - besides the better readability (in the browser) - a super argument and will definitely be very helpful in the future.

I initially failed to notice that you generated headings from the parameters. Therefore, the structure should also be preserved. 

Please be so kind and add the modified Readme.md to this PR. -- Very good work!

sachilles (approved):
LGTM

Issue #2549, opened 2022-04-28
Gitlab unstable after upgrade (14.8 to 14.10)


red-gecko27:
Hi,

When I merge old branch (previously created in v14.8) with the web interface (in v14.10) repository breaks.

When pull or clone, this error message is displayed:

```
git pull
fatal: commit-graph requires overflow generation data but has none
fatal: the remote end hung up unexpectedly
```

guangee:
> Hi,
> 
> When I merge old branch (previously created in v14.8) with the web interface (in v14.10) repository breaks.
> 
> When pull or clone, this error message is displayed:
> 
> ```
> git pull
> fatal: commit-graph requires overflow generation data but has none
> fatal: the remote end hung up unexpectedly
> ```

i meet the same error,maybe it is a bug! i guess it can broken the repository

f-io:
Temporary workaround, go to gitaly relative path: @hashed/...  and switch `git config core.commitGraph false`

stefanfritsch:
I have the same issue. A user broke a repo by doing a merge. I'd recommend everyone who still can should not upgrade to 14.10 (as gitlab doesn't support downgrades). f-io's workaround helped in our case. Perhaps some more info for people not up to speed on gitlab's internals:

1. There should be a `/home/git/data/repositories` dir in your container (which should be on a persistent volume).
2. Most of your repos will be in a subdirectory: `@hashed/`
3. To get the hash google for an online sha256 calculator. Take the project id number (visible directly below the project name in the web interface, even for broken repos), hash it
4. Now descend the directories according to the hash (i.e. the first subdir is the first two characters of the hash, and so on
5. In the final dir is a bare git repo of your project. Edit the `config` file and add `commitGraph = false` to the `[core]` section. Use `apt update && apt install nano` to get an editor if necessary

f-io:
> 3. To get the hash google for an online sha256 calculator. Take the project id number (visible directly below the project name in the web interface, even for broken repos), hash it

Or just: Menu -> Admin -> Projects -> select the broken one -> Project Info: Gitaly relative path: @hashed/...

stefanfritsch:
Thanks for the help on finding the hash, that's a lot nicer.

> You don't need to edit the config file manually, git is available.

How? `git config` tells me `fatal: not in a git directory` in bare repos.

In any case, I found git config always to be less intuitive and more error prone than just editing the config directly.

f-io:
[...]

PR #2553, opened 2022-05-03
refs #2549: Fix git installation and gitlab.yml

paolomainardi:
refs #2549

PR #2554, opened 2022-05-03
Address inconsistencies when using git (packaged with gitaly)

sachilles:
Inconsistencies (see #2549) are addressed in this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,7 +49,6 @@ RUN set -ex && \
       libpq5 zlib1g libyaml-0-2 libssl1.1 \
       libgdbm6 libreadline8 libncurses5 libffi7 \
       libxml2 libxslt1.1 libcurl4 libicu66 libre2-dev tzdata unzip libimage-exiftool-perl \
-      libcurl4-openssl-dev libexpat1-dev gettext libz-dev libssl-dev libpcre2-dev build-essential \
       libmagic1 \
  && update-locale LANG=C.UTF-8 LC_MESSAGES=POSIX \
  && locale-gen en_US.UTF-8 \
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,7 +49,7 @@ RUN set -ex && \
       libpq5 zlib1g libyaml-0-2 libssl1.1 \
       libgdbm6 libreadline8 libncurses5 libffi7 \
       libxml2 libxslt1.1 libcurl4 libicu66 libre2-dev tzdata unzip libimage-exiftool-perl \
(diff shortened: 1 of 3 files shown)


kkimurak:
same as #2553 ?

sachilles:
Yes, seems to be somehow parallel work (with the same result). ^^

kkimurak:
All tests passed. thanks for work. 
By the way I have some comment but which one will be merged?

sachilles:
Since I did remove one additional line (compared to #2553), feel free to commit to this PR. I guess @paolomainardi will understand and forgive me.

kkimurak:
I see.

sachilles:
@kkimurak I would like to test the image before merging. I hope that this will not pose any problems for us, even in the context of the recently published security release (see <https://about.gitlab.com/releases/2022/05/02/security-release-gitlab-14-10-1-released/>).

Dockerfile
```diff
@@ -49,7 +49,7 @@ RUN set -ex && \
       libpq5 zlib1g libyaml-0-2 libssl1.1 \
       libgdbm6 libreadline8 libncurses5 libffi7 \
       libxml2 libxslt1.1 libcurl4 libicu66 libre2-dev tzdata unzip libimage-exiftool-perl \
-      libcurl4-openssl-dev libexpat1-dev gettext libz-dev libssl-dev libpcre2-dev build-essential git-core \
+      libcurl4-openssl-dev libexpat1-dev gettext libz-dev libssl-dev libpcre2-dev build-essential \
```
kkimurak, line 52: These packages (added in #2553) can be moved to BUILD_DEPENDENCIES as well. This can reduce the size of the image by 200MB.

sachilles: Thanks for this hint. Do you have any other comments or hints?

kkimurak (approved):
`build-essential` contains` gcc`, `g ++`, `libc6-dev`, and` make`, so I think you can remove one of them.
Anyway LGTM.
 
I built it locally and checked it, but I couldn't reproduce the problem reported #2549, so I'm not sure if there are any other major problems.

sachilles:
@kkimurak What do you mean exactly by

> `build-essential` contains` gcc`, `g ++`, `libc6-dev`, and` make`, so I think you can remove one of them.
> Anyway LGTM.

kkimurak:
`gcc g++ make libc6-dev` can be replaced by `build-essential` for simplicity but this is out of scope and may not caeses any issues. Sorry for not clear.

PR #2558, opened 2022-05-05
Use git instead of git-core

sachilles:
This PR will use the package `git` instead of `git-core` just for building GitLab from source. At the end of the install script (i.e. `assets/build/install.sh`) the `git` package will be deleted and GitLab will use its own `git` binary.

This PR is related to #2549.


sachilles:
My locally built Docker image just contains the `git` from `gitaly` after building.

PR #2569, opened 2022-05-24
Upgrade GitLab CE to 15.0.0

sachilles:
GitLab Release: 15.0.0 (see <https://about.gitlab.com/releases/2022/05/22/gitlab-15-0-released/>)


sachilles:
I am releasing this version to make way for the security release. 

At least for me this version worked. However, I do not use all the possibilities and capabilities of GitLab.

Kiina:
Anyone has the option to enable container scanning with this release? It should be moved down to all tiers but it still shows "Available with Ultimate" on my side. But can't check if this is an gitlab issue or with the container since I don't have an instance of the official image one running.

PR #2580, opened 2022-06-04
fix condition to apply patch during build gitlab 

kkimurak:
This project applies some patches (in `assets/build/patches/*.patch`) for gitlab itself during build.  

No patch have been applied because the test `[[ -f ]]` with glob fails.  
This have been fixed by using `find | while read`.


kkimurak:
Hmm, this worked on my environment but seems not on CI...
I will turn this to draft until I can find out good alternate.

kkimurak:
fixed.

rrondeau:
did not see your PR before doing mine with another solution in 2592 :-)

PR #2581, opened 2022-06-05
Update docker command for sample Postgresql

alaahong:
As the default Postgresql version had been upgraded as version 12,  the sample command should be updating too, otherwise, we will meet permission issue during first execution.


sachilles:
Hi,

thanks for your contribution.

PR #2592, opened 2022-06-22
Fix patch presence detection

rrondeau:
This change fix the check before applying patches to the gitlab code.
Bash test does not support globbing.


rrondeau:
Oh, seems like someone got the same problem as me in #2580

sachilles:
I'll close this PR in favor of #2580

PR #2594, opened 2022-06-24
refs https://github.com/sameersbn/docker-gitlab/issues/2593: fix env default variable for terraform state

Syphon83:
Fix related issue. 

Please consider backporting it to 14.10 gitlab version. 

Thank you

PR #2579, opened 2022-06-04
Set `*:object_store:enabled` in gitlab.yml regardless the value

kkimurak:
In current code, `upload:object_store:enabled` is updated [only if `GITLAB_UPLOADS_OBJECT_STORE_ENABLED` is set to `true`](https://github.com/sameersbn/docker-gitlab/blob/1417c926360f553331f961e1177afe14cbccd615/assets/runtime/functions#L1157-L1190). It makes `puma` to die because of invalid value appears to configuration.  

This can be fixed by running `update_template GITLAB_UPLOADS_OBJECT_STORE_ENABLED` regardless of the value.

Here is a `/var/log/gitlab/supervisor/puma.log` that shows puma reporting exception.  

<details>
<summary>/var/log/gitlab/supervisor/puma.log</summary>

````sh
bundler: failed to load command: puma (/home/git/gitlab/vendor/bundle/ruby/2.7.0/bin/puma)
/home/git/gitlab/config/initializers/direct_upload_support.rb:22:in `verify!': Object storage provider '{"{\\"GITLAB_UPLOADS_OBJECT_STORE_CONNECTION_PROVIDER\\"=>nil}"=>nil}' is not supported when 'direct_upload' is used for 'lfs'. Only Google, AWS, and AzureRM are supported. (DirectUploadsValidator::ValidationError)
	from /home/git/gitlab/config/initializers/direct_upload_support.rb:45:in `block (2 levels) in <top (required)>'
	from /home/git/gitlab/config/initializers/direct_upload_support.rb:44:in `each'
	from /home/git/gitlab/config/initializers/direct_upload_support.rb:44:in `block in <top (required)>'
	from /home/git/gitlab/config/initializers/direct_upload_support.rb:37:in `tap'
	from /home/git/gitlab/config/initializers/direct_upload_support.rb:37:in `<top (required)>'
	from /home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/railties-6.1.4.7/lib/rails/engine.rb:681:in `load'
	from /home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/railties-6.1.4.7/lib/rails/engine.rb:681:in `block in load_config_initializer'
	from /home/git/gitlab/vendor/bundle/ruby/2.7.0/gems/activesupport-6.1.4.7/lib/active_support/notifications.rb:205:in `instrument'
[...]

PR #2596, opened 2022-06-25
Upgrade GitLab CE to 15.1.0

sachilles:
GitLab Release: 15.1.0 (see <https://about.gitlab.com/releases/2022/06/22/gitlab-15-1-released/>)


kkimurak:
Hmm the build fails with 

````sh
rake aborted!
ERROR: This installation of GitLab uses unsupported database names in 'config/database.yml': geo. The only supported ones are main, ci.
````

This is happen because [`geo` entry is exist on `config/database.yml.example.postgresql` of upstream source](https://gitlab.com/gitlab-org/gitlab/-/blob/4bff2971fa4640f7e9d844be4308070579eca005/config/database.yml.postgresql).  
To pass temporally, we need to delete it after [copying this as `database.yml`](https://github.com/sameersbn/docker-gitlab/blob/master/assets/build/install.sh#L187-L190) or have different approach.  

@sachilles Do you have any idea ?

sachilles:
I think the temporary workaround you suggest is the simplest solution. (However, we should include a note in the installation script).

sachilles:
We should perhaps include the configuration file in our repo below `assets/build` (e.g. as `assets/build/config/database.yml.postgresql`). 

This would then be a copy from the upstream repo, just without the `geo` entry.

kkimurak:
OK build succeed. but let me remind you what #2580 was for :)

sachilles:
Thanks, I'll check if I did a rebase or not. In any case, I guess including PRs #2579, #2580, and #2594 would be fine.

Probably, I should create an additional branch for version `15.0.x-stable`(or just `15.0.x`) the backport changes arising in version `15.1.x`. This is similar to the redmine repo (see <https://github.com/sameersbn/docker-redmine/>).

What do you think?

kkimurak:
@sachilles I find it difficult to patch past versions on multiple branches and provide a fixed version. Even if we have only one release branch, we are still providing releases with bugs and stale configs. 

Above all, I'm worried that the job will be a burden to the maintainer and that you are the only active maintainer. I apologize for not being able to make any further efforts. I can still contribute, but I can't be a maintainer.

PR #2663, opened 2022-11-16
Upgrade to GitLab CE 15.4.5

antt1995:


antt1995:
https://about.gitlab.com/releases/2022/11/14/gitlab-15-4-5-released/

antt1995:
Important notes on upgrading

GitLab 15.4.0 introduced a default [Sidekiq routing rule](https://docs.gitlab.com/ee/administration/operations/extra_sidekiq_routing.html) that routes all jobs to the default queue. For instances using [queue selectors](https://docs.gitlab.com/ee/administration/operations/extra_sidekiq_processes.html#queue-selector), this will cause [performance problems](https://gitlab.com/gitlab-com/gl-infra/scalability/-/issues/1991) as some Sidekiq processes will be idle.

    The default routing rule has been reverted in this release (15.4.5), so upgrading to this version or later will return to the previous behavior.

    If a GitLab instance now listens only to the default queue (which is not currently recommended), it will be required to add this routing rule back in gitlab.rb:

    sidekiq['routing_rules'] = [['*', 'default']]

sachilles (approved):
LGTM

Issue #2680, opened 2023-01-02
pg_dump: error: aborting because of server version mismatch


f-io:
Postgres 12 is deprecated and is scheduled for removal in GitLab 16.0.
https://docs.gitlab.com/ee/update/deprecations.html#postgresql-12-deprecated

Using this image with Postgres 13.8 (recommend/default for new installations) will end up in none working backups.  
Database dump (rake task) will fail with pg_dump missmatch.

```
pg_dump: error: server version: 13.8; pg_dump version: 12.13 (Ubuntu 12.13-1.pgdg20.04+1)
pg_dump: error: aborting because of server version mismatch
```

Due to the fact that this image use "postgresql-client-12" there is no pg_dump/restore available for 13.
https://github.com/sameersbn/docker-gitlab/blob/master/Dockerfile

Maybe we need to expose a new ENV to manage this and have both versions available or just make a breaking change?

fhaefemeier:
I have the same issue. I use psql 14 as external database and has the same issue as described here.

PR #2681, opened 2023-01-02
use the latest (13) postgresql-client libraries

f-io:
#2680

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,9 @@
 
 This file only reflects the changes that are made in this image. Please refer to the upstream GitLab [CHANGELOG](https://gitlab.com/gitlab-org/gitlab-foss/blob/master/CHANGELOG.md) for the list of changes in GitLab.
 
+**15.7.0-1**
+- postgresql: upgrade to postgresql 13
+
 **15.7.0**
 - gitlab: upgrade CE to v15.7.0
 - gitaly: upgrade to v15.7.0
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,7 @@ RUN set -ex && \
  && apt-get update \
  && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
       sudo supervisor logrotate locales curl \
-      nginx openssh-server postgresql-client-12 postgresql-contrib-12 redis-tools \
+      nginx openssh-server postgresql-client-13 postgresql-contrib-13 redis-tools \
       python3 python3-docutils nodejs yarn gettext-base graphicsmagick \
       libpq5 zlib1g libyaml-0-2 libssl1.1 \
       libgdbm6 libreadline8 libncurses5 libffi7 \


kkimurak:
I think it is not good idea to just bumping the version. We will face the same problem in the future as the required version should continue to be bumped. So it would be nice if there were fewer changes in the future.

Since we use postgresql's apt repository, if we install without specifying a major version, the latest version (currently 15.x) will be installed with `apt-get install postgresql-client`. Is that enough for us?
I just checked a combination of client (version "Ubuntu 15.1-1.pgdg20.04+1") and server (Ubuntu 14.6-1.pgdg20.04+1), and the backup was successful.

Or, how about installing corresponding version of `postgresql-client` checking server version?  
Here is a patch (see kkimurak/docker-gitlab@5f6fa91 or my [`install-pgsql-client-based-on-server-version`](https://github.com/kkimurak/docker-gitlab/tree/install-pgsql-client-based-on-server-version) branch):

````diff
diff --git a/assets/runtime/functions b/assets/runtime/functions
index 787ce5c8..6d56944d 100644
--- a/assets/runtime/functions
+++ b/assets/runtime/functions
@@ -167,11 +167,56 @@ gitlab_check_database_connection() {
   echo
 }

+gitlab_install_corresponding_database_client() {
+  echo "  Installing postgresql client to avoid version mismatch on dumping"
+  # server_version_num property is a number built from version string:
+  # https://www.postgresql.org/docs/15/libpq-status.html#LIBPQ-PQSERVERVERSION 
+  # > The result is formed by multiplying the server's major version number by 10000 and adding the minor version number. 
+  # > For example, version 10.1 will be returned as 100001, and version 11.0 will be returned as 110000. Zero is returned if the connection is bad.
+  # >
+  # > Prior to major version 10, PostgreSQL used three-part version numbers in which the first two parts together represented the major version.
+  # > For those versions, PQserverVersion uses two digits for each part;
+  # > for example version 9.1.5 will be returned as 90105, and version 9.2.0 will be returned as 90200.
+  #
+  # This difference also appends to apt package name.
+  # For example, in ubuntu:focal, postgresql-client-{8.2, 8.3, 8.4, 9.0, 9.1, 9.2, 9.3, 9.4, 9.5, 9.6, 10, 11, 12, 13, 14, 15} are available.
+  # 
+  DB_SERVER_VERSION=$(PGPASSWORD=${DB_PASS} psql -h "${DB_HOST}" -p "${DB_PORT}" -U "${DB_USER}" -d "${DB_NAME}" -Atw -c "SHOW server_version_num")
+  if [[ "${DB_SERVER_VERSION}" -eq 0 ]]; then
+    echo
+    echo "Could not retrieve database server version correctly. Aborting..."
+    return 1
+  fi
+
+  echo "    - Server version: ${DB_SERVER_VERSION}"
+
+  # Anyway, we can get major version (8, 9, 10 and so on) by dividing by 10000.
+  # DB_SERVER_VERSION_MAJOR=${DB_SERVER_VERSION%%.*}
+  DB_SERVER_VERSION_MAJOR=$((DB_SERVER_VERSION/10000))
+  DB_CLIENT_VERSION_PACKAGE_NAME=
+
+  if [[ "${DB_SERVER_VERSION_MAJOR}" -ge 10 ]]; then
+    # v10 or later: use "rought major version" as version number in package name
+    DB_CLIENT_VERSION_PACKAGE_NAME=${DB_SERVER_VERSION_MAJOR}
+  else
+    # prior to v10: convert
+    # FIXME: rought implementation
+    # It exploits the fact that there is no version such as 9.10, and it lacks versatility.
+    # divide by 100, then replace first 0 to comma
+    DB_CLIENT_VERSION_PACKAGE_NAME=$((DB_SERVER_VERSION/100))
+    DB_CLIENT_VERSION_PACKAGE_NAME=${DB_CLIENT_VERSION_PACKAGE_NAME/0/.}
+  fi
+  echo "    - Client version: ${DB_CLIENT_VERSION_PACKAGE_NAME}"
+  apt-get update && apt-get install -y postgresql-client-"${DB_CLIENT_VERSION_PACKAGE_NAME}"
+  rm -rf /var/lib/apt/lists/*
+}
+
 gitlab_configure_database() {
   echo -n "Configuring gitlab::database"

   gitlab_finalize_database_parameters
   gitlab_check_database_connection
+  gitlab_install_corresponding_database_client

   update_template ${GITLAB_DATABASE_CONFIG} \
     DB_ENCODING \
````

Anyway, thank you for your contribution.

kkimurak:
Well, I found that it is not recommended to dump using `pg_dump` whose major version is newer than the server side. Documentation says "Not guaranteed, Not impossible, requires additional tweak"

https://www.postgresql.org/docs/15/app-pgdump.html#PG-DUMP-NOTES

> Also, it is not guaranteed that pg_dump's output can be loaded into a server of an older major version — not even if the dump was taken from a server of that version. Loading a dump file into an older server may require manual editing of the dump file to remove syntax not understood by the older server. Use of the --quote-all-identifiers option is recommended in cross-version cases, as it can prevent problems arising from varying reserved-word lists in different PostgreSQL versions.

fhaefemeier:
@f-io I do :-) I use psql 14 as external database and has the same issue as described in #2681 . Manual updating the package to 14 in the running container works for me.

PR #2690, opened 2023-01-17
Upgrade to 12.10.14, part of the upgrade path to 15.6.X

skumar-jnj:
JATV-631

Following the guidelines outlined [here](https://docs.gitlab.com/ee/update/#upgrade-paths), 12.10.14 is the latest version of 12.X, before upgrade to 13.X (which requires an update to postgres).


lisbethw1130:
it seems this version is critical to everyone that needs to update from v12

PR #2699, opened 2023-01-24
Rollback Ruby from v3.x to v2.7.7

kkimurak:
This PR reverts #2675 that upgraded Ruby to v3.x that gitlab does not support yet.  

In #2675, `RUBY_VERSION` written in gitlab-omnibus-builder/docker/VERSIONS is cited as the basis for the upgrade. It was a version of Ruby used to run omnibus-gitlab's build scripts, not the Ruby used to install gems or anything into the package.  
Reference: https://gitlab.com/gitlab-org/gitlab-omnibus-builder/-/merge_requests/237#note_1126063965

See #2698 for detail.

- Roll back to 364df4c81d966cebfd6858b9b85f40afae4a4ff4 (v2.7.6 -> v2.7.7)
  - Partially revert 8386b46009b849e6a220bb0c5596d5c8a29a5f7a (v3.0.4 -> v3.0.5)
  - Revert 988cb396fb484ef4b684469916dd654965fea6d5 (v2.7.7 -> v3.0.4)

Affected releases are: v15.8.0, all v15.7.x and v15.6.x (15.6.3 or later)

Issue #2698, opened 2023-01-24
Gitlab is not ready to move to ruby v3


Daabramov:
Hi!

A couple of users have encountered a problem with JUnit running on versions 15.6.3+, there is an issue - https://gitlab.com/gitlab-org/gitlab/-/issues/385698. 
After investigating the problem, we found this - https://gitlab.com/gitlab-org/gitlab/-/issues/385698#note_1249613722

The cause of the problem: Gitlab is not yet ready to migrate to Ruby v3, but in community image this migration has already happened https://github.com/sameersbn/docker-gitlab/pull/2675

I suggest rolling back Ruby to version 2.7.7

kkimurak:
Thank you for your reporting, there were a confusion on `RUBY_VERSION` we need to refer. PR #2699 is going on.

Floyddotnet:
i can confirm, after downgrading back to ruby 2.7.7, it works now as before. thank you very much.

sachilles:
I'm going to close this ticket now. Anyway thanks for your hints.

PR #2703, opened 2023-01-25
[GitLab 15.7.5] Rollback Ruby from v3.0.4 to v2.7.7

sachilles:
This PR reverts #2675 that upgraded Ruby to v3.x that gitlab does not support yet. For Details see #2699 and #2698.


sachilles (commented):
Okay.

PR #2715, opened 2023-02-22
Update supervisord configs

kkimurak:
This PR introduces following changes:

- Clarify "user=root" in supervisord.conf
  We can see warning message "CRIT: Supervisor running as root" on container startup. This is not harmful but may confuse user.  
  We can simply prevent this warning by clarifying user in supervisord.conf.  
  We also set `user` for each programs so I believe nothing happens.
- Update sidekiq startup command : remove deprecated options
  Sidekiq is updated to v6.0 here (first contained tag: v14.4.0-ee)
  https://gitlab.com/gitlab-org/gitlab/-/merge_requests/69655
  
  In Sidekiq 6.0, command line option `-P` (set pid file) and `-L` (set log file) have been marked as deprecated.
  See https://github.com/sidekiq/sidekiq/commit/3f5b1c5
  
  Now, we can see error message in {GITLAB_LOGS_DIR}/supervisor/sidekiq.log
  like below:

  ````
  ERROR: PID file creation was removed in Sidekiq 6.0, please use a proper process supervisor to start and manage your services
  ERROR: Logfile redirection was removed in Sidekiq 6.0, Sidekiq will only log to STDOUT
  ````

  Log file is managed by supervisor's STDOUT option. PID is also managed by supervisord. So we don't need to set them. Just stop using this option.

### DIFF
--- a/assets/build/install.sh
+++ b/assets/build/install.sh
@@ -251,6 +251,10 @@ echo "UseDNS no" >> /etc/ssh/sshd_config
 # move supervisord.log file to ${GITLAB_LOG_DIR}/supervisor/
 sed -i "s|^[#]*logfile=.*|logfile=${GITLAB_LOG_DIR}/supervisor/supervisord.log ;|" /etc/supervisor/supervisord.conf
 
+# prevent confusing warning "CRIT Supervisor running as root" by clarify run as root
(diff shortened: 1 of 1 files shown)


sachilles:
Thanks!

PR #2717, opened 2023-02-22
Avoid permission error on creating config file

kkimurak:
`assets/build/config/database.yml.postgresql` is owned by root:root, and permission is rwxrw----
so the user `git` (`${GILTAB_USER}`) doesn't have a permission to copy the file.
This raises an permission error while building image locally if the docker is running on rootless environment (or the user is just a member of `docker` group).

This PR avoids this kind of issue by first copying the file as root and then changing the owner to the `GITLAB_USER`.

### DIFF
--- a/assets/build/install.sh
+++ b/assets/build/install.sh
@@ -193,7 +193,8 @@ exec_as_git cp ${GITLAB_INSTALL_DIR}/config/gitlab.yml.example ${GITLAB_INSTALL_
 # Temporary workaround, see <https://github.com/sameersbn/docker-gitlab/pull/2596>
 #
 # exec_as_git cp ${GITLAB_INSTALL_DIR}/config/database.yml.postgresql ${GITLAB_INSTALL_DIR}/config/database.yml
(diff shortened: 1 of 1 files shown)


kkimurak:
rebased on current master (15.9.3)

sachilles:
@kkimurak Sorry for the delay. Is the postioion you pointed out the only position that fails on rootless environments?

kkimurak:
@sachilles Yes I believe so, as far as I have been building image locally these years.  
This was introduced as a temporary workaround (#2596), and as far as I know it's the only access to configuration files (that copied from host to image `/etc/docker-gitlab/runtime`) happening in build.sh.  
Usually, these files are managed by using [`install_template`](https://github.com/sameersbn/docker-gitlab/blob/5450187484dd82ca8b1412fe93eca23f6e5ad1cc/assets/runtime/functions#L60-L72) by root user.  

But I have to say "I believe so", because I don't use all the features of gitlab.

kkimurak:
rebased onto 15.10.3

Issue #2603, opened 2022-07-04
docker buildx errors


killsnow:

killsnow:
Hi:
   when I use docker buildx to build arm64 image in x86 there are some errors。

docker buildx build --platform linux/arm64 -t gitlab:14.10.0 .

[+] Building 77.4s (8/14)                                                                                                                                                
 => [internal] load .dockerignore                                                                                                                                   0.1s
 => => transferring context: 128B                                                                                                                                   0.0s
 => [internal] load build definition from Dockerfile                                                                                                                0.1s
 => => transferring dockerfile: 3.71kB                                                                                                                              0.0s
 => [internal] load metadata for docker.io/library/ubuntu:focal-20220531                                                                                            7.2s
 => [auth] library/ubuntu:pull token for registry-1.docker.io                                                                                                       0.0s
 => [1/9] FROM docker.io/library/ubuntu:focal-20220531@sha256:fd92c36d3cb9b1d027c4d2a72c6bf0125da82425fc2ca37c414d4f010180dc19                                      0.0s
 => => resolve docker.io/library/ubuntu:focal-20220531@sha256:fd92c36d3cb9b1d027c4d2a72c6bf0125da82425fc2ca37c414d4f010180dc19                                      0.0s
 => [internal] load build context                                                                                                                                   0.0s
 => => transferring context: 1.85kB                                                                                                                                 0.0s
 => CACHED [2/9] RUN apt-get update  && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y     wget ca-certificates apt-transport-https gnu  0.0s
 => CANCELED [3/9] RUN set -ex &&     apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv E1DD270288B4E6030699E45FA1715D88E1DF1F24  && echo "deb http://  70.1s
WARNING: No output specified for docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load                                                                                                                                                 
error: failed to solve: Canceled: context canceled                                                                                                                       
xujinshui@xujinshuideMacBook-Pro-6083 docker-gitlab % vi Dockerfile 
xujinshui@xujinshuideMacBook-Pro-6083 docker-gitlab % docker buildx build --platform linux/arm64 -t gitlab:15.1.0 .
[+] Building 4.2s (7/13)                                                                                                                                                 
 => [internal] load build definition from Dockerfile                                                                                                                0.1s
[...]

PR #2732, opened 2023-03-17
Add support to build arm64 image

kkimurak:
This PR allow us to build arm64 image in addition to currently supported amd64.

Originally I have posted a patch to #2603 . I would love to hear your thoughts (and hopefully reports on how it works) from the community here - including whether or not we should support it. 

Also please note that this PR does not modify CI configuration - This is to ensure that this change does not break existing amd64 image builds and continues to work.

You can test this PR after applying the patch and run `docker buildx build --platform linux/arm64 .`

### DIFF
--- a/assets/build/install.sh
+++ b/assets/build/install.sh
@@ -14,6 +14,22 @@ RUBY_SRC_URL=https://cache.ruby-lang.org/pub/ruby/${RUBY_VERSION%.*}/ruby-${RUBY
 
 GEM_CACHE_DIR="${GITLAB_BUILD_DIR}/cache"
 
+# check system platform
+# TODO: cover as many variations as possible (depending on the platform supported by golang)
+case "$(uname -m)" in
+  "x86_64")
+    PLATFORM_NAME="amd64"
+    ;;
+  "aarch64")
+    PLATFORM_NAME="arm64"
+    ;;
+  *)
+    echo "The platform not supported ($(uname -a))"
+    exit 1
+esac
+
+GOLANG_ARCHIVE="go${GOLANG_VERSION}.linux-${PLATFORM_NAME}.tar.gz"
+
 GOROOT=/tmp/go
 PATH=${GOROOT}/bin:$PATH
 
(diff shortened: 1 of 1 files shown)


kkimurak:
Made sure to pass shellcheck.

kkimurak:
Corrected download message (uninitialized variable in use).

kkimurak:
rebased onto 15.10.3

kkimurak:
rebased onto 15.11.3

kkimurak:
rebased onto 16.0.3.

If this work is not merged in the future, it would be helpful if you could close it and state that there are no plans to prepare an image for the arm platform.

sachilles:
@kkimurak Apologies for my poor availability.

In principle I don't see any problems by supporting ARM64 if there are contributors that support this work an help to publish the corresponding images regularly. So far I didn't provide images for multiple architectures und thus, I'll need help on this.

Jack12816:
I would love to see this merged and made available. <3 Thanks for the contribution @kkimurak! 

Maybe [this](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/) could help here to enable a multi-arch build via CircleCI. 

I guess `--platform=linux/amd64,linux/arm64` must be added to the `docker build` commands:
* https://github.com/sameersbn/docker-gitlab/blob/master/.circleci/config.yml#L79
* https://github.com/sameersbn/docker-gitlab/blob/master/.circleci/config.yml#L105

and the `docker publish` works transparently. 

@sachilles It would be great if you could merge this PR with the CI addition and see if CI is working properly. If not revert it - then we could iterate on this. :)

kkimurak:
rebased onto 16.2.1.

gjrtimmer:
I would really love this, waiting for my RK1 modules from turing pi. With this I can run my GitLab on my Turing PI.

kkimurak:
Sorry for delay but 

- I'm not going to maintain CI settings (as I stated at first)
- I've lost motivation because the only ARM64 device I owned broke.
- There is a concurrent work "Support Arm Architecture" (#2803). This can be confusing, and I think discussions with the same content should be grouped together (i.e. one should be closed).

I hope someone who need this will take over (or #2803 - not exactly same code but works) .

I apologize again for my selfish actions.

PR #2735, opened 2023-03-24
Allow SSH clients to explicitly set the Git transfer protocol

zimmski:
This allows to the client to explicitly select protocol version 2, which is much more efficient. Without allowing that environment variable it seems that SSH connections are always stuck in version1.  See https://docs.gitlab.com/ee/administration/git_protocol.html on how to check if protocol 2 is used.


sachilles:
Thank you for your contribution.

PR #2774, opened 2023-06-02
Fix unused client removal on restarted container

### LINKED ISSUE #2773: gitlab container loops on restart without starting up properly
For any version of the container after the merge of the following changes

https://github.com/sameersbn/docker-gitlab/commit/b3b700a7019a1a2a9f17a88f81d85cd2ec8a0e67
https://github.com/sameersbn/docker-gitlab/commit/260f548c5b4bd0c4e602740128aeca6058331f4d

the container starts correctly when starting everything up the first time with `docker compose up` command but as soon as I stop the container (just the gitlab container or all), trying to restart them with `docker compose start` command, the gitlab container fails to start and enters an endless loop that continuously prints the following lines 

_gitlab-gitlab-1  Initializing logdir... 
gitlab-gitlab-1  Initializing datadir... 
gitlab-gitlab-1  Container TimeZone -> Europe/Rome 
gitlab-gitlab-1  Installing configuration templates... 
gitlab-gitlab-1  Configuring gitlab... 
gitlab-gitlab-1  Configuring gitlab::database 
gitlab-gitlab-1  Configuring /home/git/.postgresqlrc to avoid version mismatch on dumping 
gitlab-gitlab-1  - Detected server version: 120003 
gitlab-gitlab-1  - Generating /home/git/.postgresqlrc 
gitlab-gitlab-1  12 postgresql:5432 gitlabhq_production 
gitlab-gitlab-1 exited with code 1_

I made some tries with all the versions before 15.11.4 (the first version that contains that two entries) and everything worked perfectly. From that version onward we got that problem

kkimurak:
Fix #2773 , followup #2685 I have submitted

Handle the case where the target does not exist in the process of deleting unused clients.
Such a situation will not occur in newly launched container, but will occur if it is restarted.

During container restarts, container status are preserved. If the unused database client was deleted, they will continue to be removed even after restarting. So `grep` to list up unused clients will not match anything.

And, `grep` returns non-zero code on nothing match. This stops container because entrypoint sets option `-e` (exit immediately on non-zero exit code excluding some special cases)

This PR make the uninstall process to handle the case UNUSED_DB_CLIENTS is empty - just do nothing then.

### DIFF
--- a/assets/runtime/functions
+++ b/assets/runtime/functions
@@ -237,7 +237,16 @@ gitlab_uninstall_unused_database_client() {
     REGEX_DB_CLIENT_VERSIONS_IN_USE="-common${DB_CLIENT_VERSIONS_IN_USE}"
 
     # remove unused client using regex above
-    UNUSED_DB_CLIENTS=$(apt-cache pkgnames postgresql-client | grep -v -e "${REGEX_DB_CLIENT_VERSIONS_IN_USE}" | tr '\n' ' ')
+    # grep may return non-zero code on mo match, so fake the exit code with the `|| true` to swallow that
+    UNUSED_DB_CLIENTS=$(apt-cache pkgnames postgresql-client | grep -v -e "${REGEX_DB_CLIENT_VERSIONS_IN_USE}" || true)
+    if [[ "${UNUSED_DB_CLIENTS}" == "" ]]; then
+      echo "- All installed version of clients are in use. Did not uninstalled any client..."
+      return
+    fi
+
+    # just to get clean log, convert newline (package name delimiter) to single whitespace
+    UNUSED_DB_CLIENTS=$(echo ${UNUSED_DB_CLIENTS} | tr '\n' ' ')
+
     echo "- Uninstalling unused client(s): ${UNUSED_DB_CLIENTS}"
     DEBIAN_FRONTEND=noninteractive apt-get -qq -y purge -- ${UNUSED_DB_CLIENTS} >/dev/null
   fi


assets/runtime/functions
```diff
@@ -237,7 +237,16 @@ gitlab_uninstall_unused_database_client() {
     REGEX_DB_CLIENT_VERSIONS_IN_USE="-common${DB_CLIENT_VERSIONS_IN_USE}"
 
     # remove unused client using regex above
-    UNUSED_DB_CLIENTS=$(apt-cache pkgnames postgresql-client | grep -v -e "${REGEX_DB_CLIENT_VERSIONS_IN_USE}" | tr '\n' ' ')
+    # grep may return non-zero code on mo match, so fake the exit code with the `|| true` to swallow that
+    UNUSED_DB_CLIENTS=$(apt-cache pkgnames postgresql-client | grep -v -e "${REGEX_DB_CLIENT_VERSIONS_IN_USE}" || true)
```
sachilles, line 241: You don't care about newlines anymore. Does the `apt-get purge` command works in as expected?

kkimurak: Yes of course I have made sure.

And, `tr` command to convrert newline in `UNUSED_DB_CLIENTS` is moved before `purge`. Could you check [line 248](https://github.com/sameersbn/docker-gitlab/pull/2774/files#diff-934fe23ac2af75a265cbd6e8da77882402932d36cfe726f9b11f1a43d0bdc5bfR248)?

sachilles: Ooops, I missed that line. 

---

I'm glad you're safe. I hope no major damage was caused by the typhoon.

sachilles (approved):

PR #2779, opened 2023-06-07
Update (fix) postgresql-client docs section in README.md

kkimurak:
Package `postgresql-client-12` have been removed from default installed clients (#2678, especially fcd5297d5dab1f383383e5ff7826a4bda3e77f98). This PR simply apply it to README.

Also fixes some minor typo in docs added in #2685, especially b736b99.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ Configuring gitlab::database
 …
 ````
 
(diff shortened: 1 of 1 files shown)


sachilles (approved):
Thanks for pointing this out. However, exactly these sections need to be updated...

PR #2803, opened 2023-07-21
support arm architecture

zeetabit:
To be able build not only amd64 images we need to support golang downloading for proper architecture.

Example:
docker buildx build --platform=linux/arm64 -t sameersbn/gitlab .

You can even adjust (in future) your automation builds to push multiarch builds to your docker registries. This is buildkit docker feature OOTB.

### DIFF
--- a/assets/build/install.sh
+++ b/assets/build/install.sh
@@ -98,8 +98,9 @@ gem install bundler:"${BUNDLER_VERSION}"
 
 # download golang
 echo "Downloading Go ${GOLANG_VERSION}..."
-wget -cnv https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-amd64.tar.gz -P ${GITLAB_BUILD_DIR}/
-tar -xf ${GITLAB_BUILD_DIR}/go${GOLANG_VERSION}.linux-amd64.tar.gz -C /tmp/
+dpkgArch="$(dpkg --print-architecture)"
+wget -cnv https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-${dpkgArch}.tar.gz -P ${GITLAB_BUILD_DIR}/
+tar -xf ${GITLAB_BUILD_DIR}/go${GOLANG_VERSION}.linux-${dpkgArch}.tar.gz -C /tmp/
 
 # install gitlab-shell
 echo "Downloading gitlab-shell v.${GITLAB_SHELL_VERSION}..."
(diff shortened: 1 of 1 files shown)


kkimurak (commented):
Short and smart code. I like it. Thank you for your contribution. 

For your information:

1. There is a concurrent work #2732 .
2. From man page of [`dpkg-architecture(1)`](https://manpages.ubuntu.com/manpages/focal/en/man1/dpkg-architecture.1.html#examples) :
  
    > In  any  case,  you  should  never  use  dpkg  --print-architecture  to  get  architecture
    > information during a package build.
    
    What do you think about it? If you plan to use dpkg, the document seems to recommend using `dpkg-architecture -qDEB_HOST_ARCH`.

zeetabit:
Thank you for detail answer.
I fine with alternate solution - why not ;)
I don't think that it's critical to use `dpkg-architecture --print-architecture` as we are not building packages, we just downloading ready ones. We do not need to initialise them as we are not modifying them as package build steps (like in makefile at compilation steps, so on).

```
 Usage in debian/rules
       The environment variables set by dpkg-architecture are passed to
       debian/rules as make variables (see make documentation). However, you
       should not rely on them, as this breaks manual invocation of the
       script. Instead, you should always initialize them using dpkg-
       architecture with the -q option. Here are some examples, which also
       show how you can improve the cross compilation support in your package:

       Retrieving the GNU system type and forwarding it to ./configure:
```

BTW I checked only arm64/amd64 builds. If you are going to support other architectures for all binaries in Dockerfile , for sure we should check golang binaries (and other (!) binaries), for golang list is there https://storage.googleapis.com/golang/, for example I see that `i386` architecture stored without `i`, so script with `if/else` or `switch/case` blocks is needed (like at alternate solution).
For now I would say it will be overengenered, but can be implemented for sure.

zeetabit:
So I will adjust PR from
```
dpkgArch="$(dpkg-architecture --print-architecture)"
```
to
```
dpkgArch="$(dpkg-architecture -qDEB_HOST_ARCH)"
```
and that's all. It still will support arm64/amd64, but other architectures should be adjusted additionally (and other binaries, starting from basic ubuntu image architecture list support)

zeetabit:
updated as mentioned, checked by 
` docker buildx build --platform=linux/amd64 --cache-from sameersbn/gitlab -t sameersbn/gitlab .`
` docker buildx build --platform=linux/arm64 --cache-from sameersbn/gitlab -t sameersbn/gitlab .`

nohnaimer:
Hello,
You can build it like this:
```bash
~$ docker buildx create --use
~$ docker buildx build --platform=linux/amd64,linux/arm64 --cache-from sameersbn/gitlab -t sameersbn/gitlab:16.4.1 .
```
But improvements are needed for the arm, because now it gives an error:
```text
#0 176.2 go: downloading google.golang.org/appengine v1.6.7
#0 189.3 # runtime/cgo
#0 189.3 gcc: error: unrecognized command line option '-m64'
#0 239.0 make: *** [Makefile:99: bin/gitlab-shell] Error 1
------
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
Dockerfile:60
--------------------
  58 |
  59 |     COPY assets/build/ ${GITLAB_BUILD_DIR}/
  60 | >>> RUN bash ${GITLAB_BUILD_DIR}/install.sh
  61 |
  62 |     COPY assets/runtime/ ${GITLAB_RUNTIME_DIR}/
--------------------
ERROR: failed to solve: process "/bin/sh -c bash ${GITLAB_BUILD_DIR}/install.sh" did not complete successfully: exit code: 2
```

gjrtimmer:
So it looks like a need to write patches that update the CGO flags in GoLang to set the reight flags for compilation for arm64

PR #2813, opened 2023-08-05
Upgrade GitLab CE to 16.1.4

sachilles:
GitLab Patch Release: 16.1.4 (see <https://about.gitlab.com/releases/2023/08/03/gitlab-16-1-4-released/>)


Swati-Gawali:
Please share the upgrade path from Gitlab CE from 15.9.1 to 16.x.x

avvertix:
@Swati-Gawali I think the merge request is not the best location to ask for the upgrade path. It should probably be an issue. 

Anyway, following the official [Gitlab upgrade path](https://docs.gitlab.com/ee/update/index.html#upgrade-paths) probably the best scenario would be to:

- Upgrade to 15.10.x (the last available here is probably 15.10.7)
- Upgrade to 15.11.x, probably in two steps 15.11.0 and then 15.11.13 (moving to 15.11.13 is [mandatory](https://docs.gitlab.com/ee/update/index.html#upgrade-paths))
- Migrate your database to Postgres 13, as Gitlab 16 requires PostgreSQL 13 or above
- Review the [official upgrade to 16.0 guide](https://docs.gitlab.com/ee/update/versions/gitlab_16_changes.html) as might contain relevant information for you

Between all the steps you must ensure that all background migrations are complete, this is extremely important as a failure in a background migration will cause troubles. For the same reason follow what's in the readme to have a working backup every time.

After that you can probably do

- 16.0.x
- 16.1.x
- 16.2.x
- 16.3.x
- and so on

Then do an upgrade to the latest available

sachilles:
According to the [upgrade path tool](https://gitlab-com.gitlab.io/support/toolbox/upgrade-path/) the recommended upgrade path to the latest version is 15.9.1 -> 15.11.13 -> 16.1.5 -> 16.3.6 -> 16.5.1.

However, please open a separate ticket the next time.

Best regards.

PR #2816, opened 2023-08-21
Upgrade nodejs apt source repo from 16.x to 18.x

kkimurak:
Required version of node.js is updated from v16.15.0 to v18.16.0 with GitLab v16.1.0 release
- "Update required node version to 18.16.0 or later"  
  https://gitlab.com/gitlab-org/gitlab/-/merge_requests/106898

Then nodejs v18.16.1 with GitLab v16.2.0
- "Update Node.js to 18.16.1"
  https://gitlab.com/gitlab-org/gitlab/-/merge_requests/124802

And will be updated to v18.17.0 soon (not released yet, but already merged)
- "Update Node.js to 18.17.0"  
  https://gitlab.com/gitlab-org/gitlab/-/merge_requests/127560

This pull request simply change the apt source repository from node_16.x to node_18.x.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ RUN set -ex && \
  && wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
  && echo 'deb http://apt.postgresql.org/pub/repos/apt/ focal-pgdg main' > /etc/apt/sources.list.d/pgdg.list \
  && wget --quiet -O - https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - \
- && echo 'deb https://deb.nodesource.com/node_16.x focal main' > /etc/apt/sources.list.d/nodesource.list \
+ && echo 'deb https://deb.nodesource.com/node_18.x focal main' > /etc/apt/sources.list.d/nodesource.list \
  && wget --quiet -O - https://dl.yarnpkg.com/debian/pubkey.gpg  | apt-key add - \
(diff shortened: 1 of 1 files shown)


kkimurak:
TODO: add changelog entry

Suggestion: Changelog entries are release-based. A merge request directly related to a release adds a new entry for that version, but it might be useful to have an entry like "latest" that temporarily writes anything else.

kkimurak:
Hi, thank you for prompt response. 

> you think that an changelog entry, that is not directly related to a relase upgrade, should be present such that the the images tagged with "latest" and related changes would be transparent to anyone using that image. (And of course, the "rolling release tag" would live just in between releases?)

Generally yes.  

More specifically, when I wrote this comment, my suggestion was not intended for the convenience of users who use the latest tag, but for contributors who want to record their changes in the changelog and for those who check the changelog. (Not intended, but it would certainly be beneficial to people like those you mentioned.)
Changes that are not pull requests, such as release upgrades, are unclear to contributors where to list them in the changelog, since it is undecided which release they will be included in. Older changelogs show entries such as "added configuration parameter {XX}," but there are no change log entries, even though similar changes have been made more recently.

sachilles:
I'll think about your suggestion and I'll check, if there is some kind of tool to support automatically generated "release" informations for the commits starting at the latest commit and ending just before the last tag of branch `master`. (The mechanism itself is not already present, but so far we don't automatically tag versions.)

As an alternative we could also introduce some kind of note, that will allow to describe changes as some kind of pre-release changelog.

However, in favor of the last GitLab release I'll collect your PR and release this together with PR #2817.

PR #2849, opened 2023-10-24
Added missing configuration variable OAUTH_ALLOW_BYPASS_TWO_FACTOR.

massej:
This variable is important due to the fact that two authenticators are already enabled within the provider. 

Without this variable, it will prompt the user for two-factor authentication twice if GitLab enforces 2FA.


README.md
```diff
@@ -2023,6 +2023,10 @@ Allow users with existing accounts to login and auto link their account via the
 
 Comma separated list if oauth providers to disallow access to `internal` projects. Users creating accounts via these providers will have access internal projects. Accepted values are `cas3`, `github`, `bitbucket`, `gitlab`, `google_oauth2`, `facebook`, `twitter`, `saml`, `crowd`, `auth0` and `azure_oauth2`. No default.
 
+##### `OAUTH_ALLOW_BYPASS_TWO_FACTOR`
+
+Comma separated list of oauth providers where users can sign in without using two-factor authentication (2FA). Defaults to `[]`.
```
kkimurak, line 2028: If you agree to remove surrounding square brackets, this line also need to be updated.  
I'm not sure what happen if we set the default value to `[]` but I believe it works as expect (equivalent to setting `false`?)

```suggestion
Specify oauth providers where users can sign in without using two-factor authentication (2FA). You can define this using an array of providers like `["twitter", "google_oauth2"]`. Setting this to `true` or `false` applies to all - allow all or none. Defaults to `false`.
```

kkimurak (commented):
Thank you for your contribution.  
I have a comment about surrounding square brackets in gitlab.yml. Feel free to ignore me if you beleive no one want to set `true` (allow all) or `false` (allow none)

massej:
Thank you @kkimurak, it is indeed better, it give more flexbility

sachilles (approved):
LGTM

sachilles:
Thanks for your contribution.

PR #2871, opened 2023-12-24
Upgrade to GitLab CE 16.7.0

sachilles:
GitLab 16.7 Release (see <https://about.gitlab.com/releases/2023/12/21/gitlab-16-7-released/>)


sachilles:
Because of the change notes, in particular

> As of GitLab 16.7, PostgreSQL 14 is the default version installed with the Linux package. During a package upgrade, the database isn’t upgraded to PostgreSQL 14. If you want to upgrade to PostgreSQL 14, [you must do it manually](https://docs.gitlab.com/omnibus/settings/database.html#upgrade-packaged-postgresql-server).
> 
> PostgreSQL 14 isn’t supported on Geo deployments and is [planned](https://gitlab.com/groups/gitlab-org/-/epics/9065) for future releases.
> 
> If you want to use PostgreSQL 13, you must set `postgresql['version'] = 13 in /etc/gitlab/gitlab.rb`.

it might be possible to think about how to support PostgreSQL 13 temporarily.

kkimurak:
It should be a topic for Linux package installation so that we have nothing to do here (same as v15.5 release, requires postgres 12), I think.

> > t upgraded to PostgreSQL 14. If you want to upgrade to PostgreSQL 14, [you must do it manually](https://docs.gitlab.com/omnibus/settings/database.html#upgrade-packaged-postgresql-server).
> > PostgreSQL 14 isn’t supported on Geo deployments and is [planned](https://gitlab.com/groups/gitlab-org/-/epics/9065) for future releases.
> > If you want to use PostgreSQL 13, you must set `postgresql['version'] = 13 in /etc/gitlab/gitlab.rb`.
> 
> it might be possible to think about how to support PostgreSQL 13 temporarily.

sachilles:
@kkimurak Thanks for pointing this out. And of course, merry xmas.

PR #1358, opened 2017-09-24
GitLab CE 10.0.0

evenh:
This is a cleaned up version of #1357.

PR #2883, opened 2024-01-18
Remove ajv as an explicit dependency

kkimurak:
Around v10.0.0 release, there was an issue so that we have to install `ajv` explicitly. This issue have been fixed on upstream, a few years ago.

It seems that fix MR on upstream is https://gitlab.com/gitlab-org/gitlab-foss/-/merge_requests/14543 but we cannot check diff because source / target branch removed by mistake.

Anyway, no need to add ajv as an explicit dependency on (at least) gitlab v11 or later. We can revert the change.
This commit partially reverts 985d57afb9673b2f5acb1f12cbc13f230f6ec074 (pull request #1358)

### DIFF
--- a/assets/build/install.sh
+++ b/assets/build/install.sh
@@ -203,7 +203,6 @@ chown ${GITLAB_USER}: ${GITLAB_INSTALL_DIR}/config/database.yml
 
 # Installs nodejs packages required to compile webpack
 exec_as_git yarn install --production --pure-lockfile
-exec_as_git yarn add ajv@^4.0.0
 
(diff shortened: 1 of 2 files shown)


kkimurak:
Note : this PR also reduce error message on build / startup (only when relative url is used) like below:

````
[5/5] Building fresh packages...
warning "ajv" is already in "devDependencies". Please remove existing entry first before adding it to "dependencies".
````

sachilles:
@kkimurak Very nice. Thank you!

PR #2897, opened 2024-02-07
Add patch for Ruby to fix issue on running on legacy kernel

kkimurak:
Partially fix #2872 . 

This PR contains following changes:

- Revise build patch structure to allow patching ruby  
  Currently we can apply some patches only for gitlab itself. This change changes the build patch folder tree and makes it similar to runtime/config, allowing build patches to be applied to both gitlab and ruby.
- Add patch to fix behavior of `SecureRandom.gen_random_openssl(n)` (See https://github.com/ruby/ruby/commit/64e503eb62aff0952b655e9a86217e355f786146).   
This commit is originally merged into upstream (Ruby 3.3.x) but can be applied to 3.1.x or later, and neccessary to run it on legacy kernel (For example, the one used by Synology NAS that claims to be able to operate `sameersbn/gitlab`)

PR #2898, opened 2024-02-07
Backport to 16.7.x : Add patch for Ruby to fix issue on running on legacy kernel 

kkimurak:
This PR is backport of #2897 to 16.7.x branch.

This PR contains following changes:

- Revise build patch structure to allow patching ruby  
  Currently we can apply some patches only for gitlab itself. This change changes the build patch folder tree and makes it similar to runtime/config, allowing build patches to be applied to both gitlab and ruby.
- Add patch to fix behavior of `SecureRandom.gen_random_openssl(n)` (See https://github.com/ruby/ruby/commit/64e503eb62aff0952b655e9a86217e355f786146).   
This commit is originally merged into upstream (Ruby 3.3.x) but can be applied to 3.1.x or later, and neccessary to run it on legacy kernel (For example, the one used by Synology NAS that claims to be able to operate `sameersbn/gitlab`)


sachilles:
@kkimurak Could you be so kind and fix the merge conflict? (Did I caused the conflict?)

kkimurak:
@sachilles Sorry for late, conflictions have been fixed. #2895 modified one of patch and this PR tried to move the patch, then caused confliction. I should have written a patch so that we could merge them all at once..

PR #2936, opened 2024-04-29
Update redis image to use latest patch and not a hardcoded patch

froque:
Current latest patch is 6.2.14


sachilles (approved):
LGTM

sachilles:
@froque Thank you for pointing out and your contribution!

Issue #2962, opened 2024-06-28
CI failure in github release due to go version of executor


kkimurak:
(search work in progress)

We use [cimg/go:1.14] for github-release but it does not compatible with recent release. Maybe we should update the image tag

- release workflow installs `meterup/github-release` by executing `go get github.com/meterup/github-release`
- `meterup/github-release` depends on `kevinburke/rest`
  - it now uses `io.ReadAll` that is added on go 1.16.
    This change have been made by the commit kevinburke/rest@ac09c3f0ec455f78bda023264dcd0857e91a6dcd that is included since `kevinburke/rest@2.7`  
    I'm not sure which version `meterup/github-release` depends on

cc: @sachilles

kkimurak:
1.16 : not sure it works - is the final line non-critical warning?

```log
go: downloading github.com/meterup/github-release v0.7.5
go: downloading github.com/dustin/go-humanize v1.0.1
go: downloading github.com/voxelbrain/goptions v0.0.0-20180630082107-58cddc247ea2
go: downloading github.com/kevinburke/rest v0.0.0-20240617045629-3ed0ad3487f0
go: downloading github.com/tomnomnom/linkheader v0.0.0-20180905144013-02ca5825eb80
go: downloading github.com/inconshreveable/log15 v3.0.0-testing.5+incompatible
go: downloading github.com/mattn/go-colorable v0.1.13
go: downloading github.com/go-stack/stack v1.8.1
go: downloading golang.org/x/term v0.13.0
go: downloading github.com/mattn/go-isatty v0.0.16
go: downloading golang.org/x/sys v0.14.0
../go/pkg/mod/github.com/mattn/go-isatty@v0.0.16/isatty_tcgets.go:7:8: //go:build comment without // +build comment
```

kkimurak:
1.17: may works but command should be updated to use `go install` instead of `go get` as it have been deprecated.  

```log
go: downloading github.com/meterup/github-release v0.7.5
go: downloading github.com/voxelbrain/goptions v0.0.0-20180630082107-58cddc247ea2
go: downloading github.com/tomnomnom/linkheader v0.0.0-20180905144013-02ca5825eb80
go: downloading github.com/kevinburke/rest v0.0.0-20240617045629-3ed0ad3487f0
go: downloading github.com/dustin/go-humanize v1.0.1
go: downloading github.com/inconshreveable/log15 v3.0.0-testing.5+incompatible
[...]

PR #2963, opened 2024-06-28
CI: bump jobs.release.executor to go/default:1.22

kkimurak:
See #2962 for details.

Currently CI job `release` is failing on `go get github.com/meterup/github-release` because one of the dependencies uses a feature that will be introduced in a future version of go.  
See log: https://app.circleci.com/jobs/github/sameersbn/docker-gitlab/7095

This PR bumps go image version from 1.14 to 1.22 and update command to use `go install` instead of `go get`. As `go install` requires a version when current directory is not in amodule, this PR just use `@latest`. If we need to fix the behavior, we'd better to change it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -269,12 +269,12 @@ jobs:
   release:
     executor:
       name: go/default
-      tag: '1.14'
+      tag: '1.22'
     steps:
       - checkout
       - run:
           name: Installing github-release tool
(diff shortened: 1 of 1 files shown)


sachilles:
@kkimurak Thank you for the very quick analysis and for providing a workaround. I think it is important that the release information is published again first.

However, we should change the installation source as soon as possible.

Issue #2526, opened 2022-03-20
using IAM profile for AWS backups


jasonyihk:
when `AWS_BACKUP_USE_IAM_PROFILE: true` set in the environment to use IAM profile for AWS backups, the code still using `AWS_BACKUP_ACCESS_KEY_ID` & `AWS_BACKUP_SECRET_ACCESS_KEY` to upload to S3. Also, `AWS_BACKUP_ACCESS_KEY_ID`  &`AWS_BACKUP_SECRET_ACCESS_KEY`  can not be set to empty or not being set in the enviroment.

PR #3016, opened 2024-09-29
feat: Enable IAM instance profile (IAM role) when used in AWS

ntnamazu:
GitLab have supported IAM instance profile (IAM role) functionality as stated in official documentations like this:

- [Back up GitLab](https://docs.gitlab.com/ee/administration/backup_restore/backup_gitlab.html)
- [IAM roles for AWS when using the GitLab chart](https://docs.gitlab.com/charts/advanced/external-object-storage/aws-iam-roles.html)

The code of this Pull Request enables to address the feature above. If this feature become available, dockerized GitLab booted on AWS will not require secret access key, so users can eliminate the hassle of managing access keys.

I guess this funciontality is very useful for users deploying dockerized GitLab in AWS. I know there have been a request for this feature like Issue #2526, and if the code of this Pull Request is merged, I guess the issue #2526 can be closed.

This feature was something I wanted very much myself, so I guess some of the other users will find it useful.

I would be happy if this code reviewed.


Dockerfile
```diff
@@ -82,6 +82,8 @@ LABEL \
 
 EXPOSE 22/tcp 80/tcp 443/tcp
 
+RUN ln -s /etc/ssl/certs/ca-certificates.crt /usr/lib/ssl/cert.pem
+
```
kkimurak, line 86: Sorry for late comment but this line was submitted as a separate pull request, #3027, which has been merged (and removed again in #3107, as it is no longer needed in Ubuntu Noble).
```suggestion
```

kkimurak (changes requested):
Generally looks good to me.

Issue #3010, opened 2024-09-25
AWS S3 Backup fails with SSL_CTX_load_verify_locations


fhaefemeier:
I configured automatic backups with storing the backup archive in AWS S3 (self hosted minio cluster). After upgrade the docker image from 17.1.1 to 17.3.2 the backup configuration fails with a SSL error `SSL_CTX_load_verify_locations`. The configuration is not changed. Server certificate of minio cluster is a Let's Encrypt certificate.

Configuration:
```
GITLAB_BACKUP_SCHEDULE: daily
GITLAB_BACKUP_TIME: "23:00"
GITLAB_BACKUP_SKIP: "uploads,artifacts,packages,builds,lfs,registry,ci_secure_files,terraform_state"
GITLAB_BACKUP_EXPIRY: 302400
AWS_BACKUPS: "true"
AWS_BACKUP_REGION: "de-local-1"
AWS_BACKUP_BUCKET: "gitlab-backup"
AWS_BACKUP_ACCESS_KEY_ID: <masked>
AWS_BACKUP_SECRET_ACCESS_KEY: <masked>
AWS_BACKUP_ENDPOINT: https://main.storage.<domain>
```

Log extract:
```
2024-09-23 23:00:33 +0200 -- Creating backup archive: 1727125223_2024_09_23_17.3.2_gitlab_backup.tar ... 
rake aborted!
Excon::Error::Socket: SSL_CTX_load_verify_locations: system lib (OpenSSL::SSL::SSLError)
/home/git/gitlab/vendor/bundle/ruby/3.2.0/gems/excon-0.99.0/lib/excon/ssl_socket.rb:139:in `initialize'
/home/git/gitlab/vendor/bundle/ruby/3.2.0/gems/excon-0.99.0/lib/excon/ssl_socket.rb:139:in `new'
/home/git/gitlab/vendor/bundle/ruby/3.2.0/gems/excon-0.99.0/lib/excon/ssl_socket.rb:139:in `initialize'
/home/git/gitlab/vendor/bundle/ruby/3.2.0/gems/excon-0.99.0/lib/excon/connection.rb:474:in `new'
/home/git/gitlab/vendor/bundle/ruby/3.2.0/gems/excon-0.99.0/lib/excon/connection.rb:474:in `socket'
/home/git/gitlab/vendor/bundle/ruby/3.2.0/gems/excon-0.99.0/lib/excon/connection.rb:121:in `request_call'
[...]

PR #3022, opened 2024-10-10
[GitLab Critical Patch Release] Upgrade GitLab CE to 17.3.5

sachilles:
GitLab Critical Patch Release: 17.3.5 (see <https://about.gitlab.com/releases/2024/10/09/patch-release-gitlab-17-4-2-released/>)


ymazzer:
Hi @sachilles,

Does the ubuntu base image update also fixes the issue #3010.

If not do you think it's possible to at least cherry-pick this [commit](https://github.com/sameersbn/docker-gitlab/commit/d8d7804a1d1fdd69c1fb5c4f61e79e31e8743333) to revert the regression?

kkimurak:
@ymazzer The issue #3010 will not be fixed by updating base image to newer ubuntu:focal (#3023). It still uses openssl 1.x as far as I know.

PR #3027, opened 2024-10-24
fix(#3010) Add symbolic link to the correct location of ca-cert file

ymazzer:
A pull request including the fix for the #3010 issue as fixed in the PR #3016 but isolated to make it easier to merge.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -82,6 +82,8 @@ LABEL \
 
 EXPOSE 22/tcp 80/tcp 443/tcp
 
+RUN ln -s /etc/ssl/certs/ca-certificates.crt /usr/lib/ssl/cert.pem
+
 VOLUME ["${GITLAB_DATA_DIR}", "${GITLAB_LOG_DIR}","${GITLAB_HOME}/gitlab/node_modules"]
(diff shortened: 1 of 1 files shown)


ymazzer:
@sachilles this issue is quite blocking, is it possible for you to include it in the coming update for 17.5.1?

sachilles (approved):
Looks good to me

sachilles:
@ymazzer Of course, I'll merge your PR once the CI finished correctly. The problem an ist solution seems to bi quite general. Do you think that we should backport your PR to the last 2 supported versions (i.e. 17.4.x and 17.3.x)?

ymazzer:
> @ymazzer Of course, I'll merge your PR once the CI finished correctly. The problem an ist solution seems to bi quite general. Do you think that we should backport your PR to the last 2 supported versions (i.e. 17.4.x and 17.3.x)?

@sachilles at least 17.4.x, since the problem was present in 17.4.x, I didn't tested it in 17.3.x.

This is a hack to fix the issue, btw this has to be reworked correctly for the coming 17.7 with openssl3

PR #3029, opened 2024-10-25
Upgrade GitLab CE to 17.4.3

sachilles:
GitLab Patch Release: 17.4.3 (see <https://about.gitlab.com/releases/2024/10/23/patch-release-gitlab-17-5-1-released/>)


pierreozoux:
@sachilles do you have a script to PR a new release?

Then we could also do it as well :)

There is a new cve that is fixed in 17.4.5

sachilles:
@pierreozoux Currently not and I double-check all used dependencies.

PR #3043, opened 2024-12-15
increase max-old-space-size

th-2021:
Setting --max-old-space-size to 8192 decreased my build time. Maybe it's enough for 17.6


th-2021:
Idea didn't work

Issue #3042, opened 2024-12-10
Build 17.6.0 timed out


lekoder:
CI build for 17.6.0 seems to have timed out and 17.6.0 is not available on `dockerhub`. Please run the task again.

kkimurak:
duplicate of #3036 and meaningless suggestion.

The build time of gitlab itself is getting longer and longer, and will almost certainly reach 1 hour (the upper limit in the free plan). If we don't review the build process itself, we will just waste time running CI.

lekoder:
Apologies, search didn't pick that up, since the specific version is not mentioned in the issue.

Might I suggest a workaround then of building it in another environment and pushing to the dockerhub?

kkimurak:
> Might I suggest a workaround then of building it in another environment and pushing to the dockerhub?

I'm running a locally built version of 17.6.0, but if someone publishes it on dockerhub it will help people who can't build it themselves.  
On the other hand, I'm just a contributor. You should ask @sachilles, the maintainer who has the final words.

lekoder:
Fair enough. In the meanwhile, if anybody needs this very image built to 17.6.0 up to 17.6.2 I forked it, built them and published on [the dockerhub here](https://hub.docker.com/r/koder/gitlab/tags).

th-2021:
I'm working on a multistage Dockerfile. Using BuildKit (docker buildx build or DOCKER_BUILDKIT=1 docker build), which allows parallel execution of stages I could save around 5 minutes.

sachilles:
@kkimurak Unfortunately, I don't have the option of giving you the role of maintainer.

@sameersbn  Do you see a possibility to either change the plan for the CI or take up kkimurak's offer. He has also maintained the project remarkably so far.

tDo:
@sachilles @kkimurak given that @sameersbn does not seem to be very active on github anymore (At least there have been no contributions in the last year) would moving the ci/cd + docker-image to a new namespace controlled by either of you be a feasible option?

Thomas-Ganter:
Just for curiosity’s sake: What *are* the requirements to build the image? 
I got errors 137 after 2 hours and after increasing memory to 6GB error 134 during the build after 3.5 hours — I found no info in any "building yourself" docu (or was merely looking in the wrong places). 

Any insights?

kkimurak:
@tDo If necessary, we should do so. I would like to get permission from all maintainers (with respect) if possible, but if we can't contact them, there is nothing we can do.  
[...]

PR #3045, opened 2024-12-17
Upgrade to 17.6.2

Patrick-Remy:
GitLab Patch Release: 17.6.2 (see https://about.gitlab.com/releases/2024/12/11/patch-release-gitlab-17-6-2-released/)


kkimurak:
Note: We are temporarily suspending our release process. This is because the build time exceeds the 1 hour limit of CircleCI's free plan.
It is possible to merge your patches, but the release will not happen automatically.

Also, please do not mix multiple upgrade to single pull request.

Patrick-Remy:
Hope the investigation for the long build-times will be successful, as building myself also takes about 45min :-) 

Okay, I added both commits so that anyone who want to build itself, can also build the subsequent version.
But if it's against the convention, you could also squash them on merge.

Oicho:
Regarding the circle ci issue I could lend some compute power and install what is needed for the integration with circle ci.

Looks like self hosted runners are allowed in the free tier. https://circleci.com/docs/plan-free/#self-hosted-runners

Feel free to reach out.

kkimurak:
@Oicho Really Thank you for the very attractive offer, but there's still a problem with that. As far as I know, none of the currently active maintainer have access to CircleCI's settings. Only users with Admin role can register Self Hosted Runners. Probably @sachilles has contributor privileges since it seems to have no privilege to manage role (correct me if I'm wrong).  
Circleci docs: 
- permission table : https://circleci.com/docs/ja/roles-and-permissions-overview/
- role management : https://circleci.com/docs/ja/manage-roles-and-permissions/

Furthermore, it seems that these settings are applied to each organization, not per project. It makes the current setup where everything is in the sameersbn namespace is even more inappropriate (in the sense that it is difficult to maintain without sameersbn).

Collatum-IT-GmbH-Admin:
Is there a timeline when we can expect a new gitlab version? Or is the build issue still unresolved?

kkimurak:
> Is there a timeline when we can expect a new gitlab version? 

No 

> Or is the build issue still unresolved?

Yes, maybe merging #3049 can temporarily solve the build time issue.  

Please don't rush. If you are in a hurry, build it yourself or use the pre-release provided by a kind community member: https://github.com/sameersbn/docker-gitlab/issues/3042#issuecomment-2536881865

kkimurak:
Also note that this patch, which skips the intermediate version, will probably never be directly merged. Given the history, most of the time when such a pull request has been created, it has been reworked by the maintainer into separate pull requests, each with its own release. This is because there have been many people who have complained about intermediate versions not being released.

sachilles:
Closed in favor of #3053 . Anyway thanks for your contribution, that "lives in #3053".

PR #3048, opened 2025-01-10
Docker Multistage for parallel build

th-2021:
This PR moves the builds from install.sh into Dockerfile.multistage, allowing for parallel build execution (with docker buildkit).

This changes circleci configuration to use large modell (4 CPU, 15 GB memory). Also buildkit is enabled. This brings down build time to 48 min on circleci.

@sachilles and @kkimurak please review this.

The change is parallel to the existing files: Dockerfile.multistage and install2.sh are used.


kkimurak:
Ok, it's done to build. It takes about 55m in total (47m to build).
  
First impression: Is it possible to separate multistage build and circleci config updates?  
As far as I read the comments in the PR, I can see that just specifying "resource_class: large" and "use-buildkit: true" to circleci config will reduce the build time to less than an hour (I have never tested so I'll try later by myself..). If that's the case, we should merge this first. If we only need to make a few changes, the review time will be shorter and we will be able to start the release process quickly.

Regarding the multistage build, there are a lot of changes, and another comment  pointed out that the image size has increased, so I think it would be a good idea to take some time to investigate this.

th-2021:
Ok. Build worked within the same time range ~48 min, I'll separate circleci into a new PR.

kkimurak:
Great!

PR #3053, opened 2025-01-18
Upgrade GitLab CE to 17.6.2

sachilles:
GitLab Patch Release: 17.6.2 (see <https://about.gitlab.com/releases/2024/12/11/patch-release-gitlab-17-6-2-released/>)


sachilles:
This is used instead of PR #3045. However, this PR partially contains commits of #3045 (transferred via `git cherrypick`).

PR #3059, opened 2025-01-21
Upgrade GitLab CE to 17.8.0

sachilles:
GitLab 17.8 Release (see <https://about.gitlab.com/releases/2025/01/16/gitlab-17-8-released/>)


sachilles:
**Todo's**

- [ ] check if we need to activate GitLab agent server for Kubernetes (KAS) optionally; for more informations see <https://docs.gitlab.com/ee/update/versions/gitlab_17_changes.html#1780>

th-2021:
Just found: Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8))
jammy is last release with apt-key

kkimurak:
@th-2021 kkimurak/docker-gitlab@46aa38e should fixes it, I faced an issue that local build stucks on executing apt-key and found that replacing apt-key solved the issue, so I think it might works

sachilles:
@kkimurak and @th-2021 Thanks for your hint and fix.

kkimurak:
@sachilles As far as I remember it built successfully but I could not register an agent. So we cannot include it, at least not in this release.

Issue #3096, opened 2025-04-18
PostgreSQL 16 needed for GitLab 18


mihelm:
The blog post on the release of GitLab 17.11 contains this bit:

> In GitLab 18.0, the minimum-supported version of PostgreSQL will be version 16. To prepare for this change, on instances that don’t use [PostgreSQL Cluster](https://docs.gitlab.com/administration/postgresql/replication_and_failover/), upgrades to GitLab 17.11 will attempt to automatically upgrade PostgreSQL to version 16.

What are the prospects for a `sameersbn/postgresql` image with version 16?

pmoralesp:
Just updated to 17.11.0 and broke my setup because of this

th-2021:
The upcoming release page mentions: 18.0 (2025-05-15)

kkimurak:
I think this should probably be discussed in the sameersbn/docker-postgresql repository, not here (I don't know if there will be a response from the maintainers early on, i.e. before the release of gitlab 18 when postgresql16 becomes mandatory..)

Anyway, as a temporary workaround, I pushed `kkimurak/sameersbn-postgresql` I built myself. Currently `:15` (`:15.12`, `:15.12-1`) and `:16` (`:16.8`, `:16.8-1`) are available. I have confirmed that the scheduled backup using `:16.8-1` works properly in combination with sameersbn/gitlab:17.11.0.  
See https://hub.docker.com/r/kkimurak/sameersbn-postgresql/tags

mihelm:
@kkimurak Thanks a lot for posting that! With that I was able to upgrade, and the database migration went without a hitch.

kkimurak:
@mihelm Glad it works.  

----

For now, I plan to continue regular updates until the upstream repository is updated, I get tired of it, or I find a new place to migrate to. I might try linking it to an automatic dependency update service such as renovate, or setting up a github action to push periodically, while also learning about it. Of course, there are no guarantees.  
Also, I haven't pushed to quay.io yet. I haven't signed up for the lowest development plan yet because it costs $15/month. I haven't decided whether to sign up or not.

th-2021:
@kkimurak which tags should be used for your psql? I have 16.8-1-20250424, which works well, but I saw that you also pushed 16.8-1 without date.

kkimurak:
[...]

PR #3107, opened 2025-05-15
Upgrade GitLab CE to 18.0.0

sachilles:
GitLab 18.0 Release (for more details see <https://about.gitlab.com/releases/2025/05/15/gitlab-18-0-released/>)


sachilles:
**Hint:** Currently, the container does not start correctly. Further checks are necessary.

sachilles:
@kkimurak Just followed the discussion in #3096. Can we temporarily use your docker images for the circle-ci pipelines and for the dockerfiles in this project?

kkimurak:
@sachilles Yes of course.

> Can we temporarily use your docker images for the circle-ci pipelines and for the dockerfiles in this project?

sachilles:
> @sachilles Yes of course.
> 
> > Can we temporarily use your docker images for the circle-ci pipelines and for the dockerfiles in this project?

Thank you very much!

sachilles:
**Error on startup of the microservice**

```
WARN[0000] /home/circleci/project/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
redis-1  | 1:C 17 May 2025 07:41:45.323 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
redis-1  | 1:C 17 May 2025 07:41:45.323 # Redis version=6.2.18, bits=64, commit=00000000, modified=0, pid=1, just started
redis-1  | 1:C 17 May 2025 07:41:45.323 # Configuration loaded
redis-1  | 1:M 17 May 2025 07:41:45.324 # Server initialized
redis-1  | 1:M 17 May 2025 07:41:45.324 # WARNING Memory overcommit must be enabled! Without it, a background save or replication may fail under low memory condition. Being disabled, it can can also cause failures without low memory condition, see https://github.com/jemalloc/jemalloc/issues/1328. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
postgresql-1  | Initializing datadir...
postgresql-1  | Initializing certdir...
postgresql-1  | Initializing logdir...
postgresql-1  | Initializing rundir...
postgresql-1  | Setting resolv.conf ACLs...
postgresql-1  | Initializing database...
postgresql-1  | Configuring hot standby...
postgresql-1  | ‣ Setting postgresql.conf parameter: wal_level = 'hot_standby'
postgresql-1  | ‣ Setting postgresql.conf parameter: max_wal_senders = '16'
postgresql-1  | ‣ Setting postgresql.conf parameter: checkpoint_segments = '8'
postgresql-1  | ‣ Setting postgresql.conf parameter: wal_keep_segments = '32'
postgresql-1  | ‣ Setting postgresql.conf parameter: hot_standby = 'on'
postgresql-1  | ‣ Setting postgresql.conf parameter: data_directory = '/var/lib/postgresql/16/main'
postgresql-1  | ‣ Setting postgresql.conf parameter: log_directory = '/var/log/postgresql'
postgresql-1  | ‣ Setting postgresql.conf parameter: log_filename = 'postgresql-16-main.log'
postgresql-1  | ‣ Setting postgresql.conf parameter: ssl = 'off'
postgresql-1  | Creating database user: gitlab
postgresql-1  | Creating database: gitlabhq_production...
postgresql-1  | ‣ Loading pg_trgm extension...
postgresql-1  | ‣ Loading btree_gist extension...
postgresql-1  | ‣ Granting access to gitlab user...
postgresql-1  | Starting PostgreSQL 16...
postgresql-1  | 2025-05-17 07:41:47.199 UTC [1] LOG:  starting PostgreSQL 16.9 (Ubuntu 16.9-1.pgdg22.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0, 64-bit
postgresql-1  | 2025-05-17 07:41:47.199 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
postgresql-1  | 2025-05-17 07:41:47.199 UTC [1] LOG:  listening on IPv6 address "::", port 5432
postgresql-1  | 2025-05-17 07:41:47.205 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
postgresql-1  | 2025-05-17 07:41:47.215 UTC [159] LOG:  database system was shut down at 2025-05-17 07:41:47 UTC
postgresql-1  | 2025-05-17 07:41:47.224 UTC [1] LOG:  database system is ready to accept connections
postgresql-1  | 2025-05-17 07:42:20.212 UTC [167] ERROR:  relation "application_settings" does not exist at character 523
postgresql-1  | 2025-05-17 07:42:20.212 UTC [167] STATEMENT:  SELECT a.attname, format_type(a.atttypid, a.atttypmod),
postgresql-1  |                pg_get_expr(d.adbin, d.adrelid), a.attnotnull, a.atttypid, a.atttypmod,
postgresql-1  |                c.collname, col_description(a.attrelid, a.attnum) AS comment,
postgresql-1  |                attidentity AS identity,
postgresql-1  |                attgenerated as attgenerated
postgresql-1  |           FROM pg_attribute a
postgresql-1  |           LEFT JOIN pg_attrdef d ON a.attrelid = d.adrelid AND a.attnum = d.adnum
postgresql-1  |           LEFT JOIN pg_type t ON a.atttypid = t.oid
postgresql-1  |           LEFT JOIN pg_collation c ON a.attcollation = c.oid AND a.attcollation <> t.typcollation
postgresql-1  |          WHERE a.attrelid = '"application_settings"'::regclass
postgresql-1  |            AND a.attnum > 0 AND NOT a.attisdropped
postgresql-1  |          ORDER BY a.attnum
postgresql-1  | 
```

kkimurak:
Hmm, not improved:

https://app.circleci.com/pipelines/github/sameersbn/docker-gitlab/1927/workflows/7e70dd78-6699-41c5-b2c1-fda346a7c8aa/jobs/8426?invite=true#step-108-32771_102

> gitlab-1      | psql:/home/git/gitlab/db/structure.sql:24: ERROR:  permission denied for schema public
> gitlab-1      | rake aborted!
> gitlab-1      | failed to execute:
> gitlab-1      | psql --set ON_ERROR_STOP=1 --quiet --no-psqlrc --output /dev/null --file /home/git/gitlab/db/structure.sql --single-transaction gitlabhq_production

[db/structure.sql](https://gitlab.com/gitlab-org/gitlab/-/blob/v18.0.0-ee/db/structure.sql?ref_type=tags):24 is just an end of function. The line `CREATE EXTENSION` seems to be executed (logs `extension already exists` so it seems it failed to create function.

```sql
# db/structure.sql line 1 to 24 on v18.0.0-ee
CREATE SCHEMA gitlab_partitions_dynamic;

COMMENT ON SCHEMA gitlab_partitions_dynamic IS 'Schema to hold partitions managed dynamically from the application, e.g. for time space partitioning.';

CREATE SCHEMA gitlab_partitions_static;

COMMENT ON SCHEMA gitlab_partitions_static IS 'Schema to hold static partitions, e.g. for hash partitioning';

CREATE EXTENSION IF NOT EXISTS btree_gist;

CREATE EXTENSION IF NOT EXISTS pg_trgm;

CREATE FUNCTION assign_ci_runner_machines_id_value() RETURNS trigger
    LANGUAGE plpgsql
    AS $$
BEGIN
IF NEW."id" IS NOT NULL THEN
  RAISE WARNING 'Manually assigning ids is not allowed, the value will be ignored';
END IF;
NEW."id" := nextval('ci_runner_machines_id_seq'::regclass);
RETURN NEW;

END
$$;
```

sachilles:
I guess it's a problem with the privileges within the database; only the user `postgres` has the privileges to set some grants for the user `gitlab` on the database.

kkimurak:
I'm testing locally built image. It's a strange bit but I noticed that there is no table `information_schema.role_schema_grants` even after `GRANT ..` query. so I cannot verify if it successfully done

```
$ docker exec -it postgresql psql -h localhost -U gitlab postgres

postgres-> \l
                                                        List of databases
        Name         |  Owner   | Encoding | Locale Provider | Collate |  Ctype  | ICU Locale | ICU Rules |   Access privileges
---------------------+----------+----------+-----------------+---------+---------+------------+-----------+-----------------------
 gitlabhq_production | postgres | UTF8     | libc            | C.UTF-8 | C.UTF-8 |            |           | =Tc/postgres         +
                     |          |          |                 |         |         |            |           | postgres=CTc/postgres+
                     |          |          |                 |         |         |            |           | gitlab=CTc/postgres
 postgres            | postgres | UTF8     | libc            | C.UTF-8 | C.UTF-8 |            |           |
 template0           | postgres | UTF8     | libc            | C.UTF-8 | C.UTF-8 |            |           | =c/postgres          +
                     |          |          |                 |         |         |            |           | postgres=CTc/postgres
 template1           | postgres | UTF8     | libc            | C.UTF-8 | C.UTF-8 |            |           | =c/postgres          +
                     |          |          |                 |         |         |            |           | postgres=CTc/postgres
(4 rows)

postgres-> GRANT ALL ON SCHEMA public TO gitlab

postgres-> SELECLT * FROM information_schema.role_ # <- here I hit tab and following lines are the candidates displayed by tab completion
information_schema.role_column_grants   information_schema.role_table_grants    information_schema.role_usage_grants
information_schema.role_routine_grants  information_schema.role_udt_grants

**Comment by sachilles** [comment]
Manuall fixing the grants to

```
        Name         |  Owner   | Encoding | Locale Provider | Collate |  Ctype  | ICU Locale | ICU Rules |   Access privileges   
---------------------+----------+----------+-----------------+---------+---------+------------+-----------+-----------------------
 gitlabhq_production | gitlab   | UTF8     | libc            | C.UTF-8 | C.UTF-8 |            |           | =Tc/gitlab           +
                     |          |          |                 |         |         |            |           | gitlab=CTc/gitlab
 postgres            | postgres | UTF8     | libc            | C.UTF-8 | C.UTF-8 |            |           | 
 template0           | postgres | UTF8     | libc            | C.UTF-8 | C.UTF-8 |            |           | =c/postgres          +
                     |          |          |                 |         |         |            |           | postgres=CTc/postgres
 template1           | postgres | UTF8     | libc            | C.UTF-8 | C.UTF-8 |            |           | =c/postgres          +
                     |          |          |                 |         |         |            |           | postgres=CTc/postgres
(4 rows)

```
helps.

**Comment by sachilles** [comment]
@kkimurak Could you be so kind and check the grants mentioned in https://github.com/kkimurak/docker-postgresql?tab=readme-ov-file#granting-user-access-to-a-database ?

**Comment by sachilles** [comment]
I guess the `ALTER TABLE database_name OWNER TO database_user;` statement will work. (In our case this reads as  `ALTER TABLE gitlabhq_production OWNER TO gitlab;`.)

**Comment by th-2021** [comment]
Should the grant be added to postgresql create_database*( in https://github.com/sameersbn/docker-postgresql/blob/master/runtime/functions ?
There is a grant statement for the database itself.

**Comment by sachilles** [comment]
This is strange, since at the end of https://github.com/kkimurak/docker-postgresql/blob/5c9eda05fc162eb8370448697ad5a14b5f42ac37/runtime/functions#L344 all grants are should be given to the user `gitlab`. But the observed database owner was postgres.

**Comment by th-2021** [comment]
This is the database grant. We need also a grant to schema public.

**Comment by th-2021** [comment]
GRANT ALL ON SCHEMA public TO gitlab;

**Comment by kkimurak** [comment]
Made sure to work and pushed kkimurak/docker-postgresql@1697c4081c9ffd4ad621564134d81cdbcbd5e908. Preparing to release images (maybe `15` (`15-20250518`) and `16`(`16-20250518`))..

**Comment by kkimurak** [comment]
pushed `kkimurak/sameersbn-postgresql` tags : `:15` (15-20250518, 15.13, 15.13-1), `:16` (16-20250518, 16.9, 16.9-1) and `:17` (17-20250518, 17.5, 17.5-1)

**Comment by kkimurak** [comment]
OK test passed.

**Comment by sachilles** [comment]
Yes, tests are passed. However, on my last test on a different setup (using an external PostgreSQL 17 database) GitLab did not started correctly. As far as I remember, some components of the image died (as you can see here https://app.circleci.com/pipelines/github/sameersbn/docker-gitlab/1930/workflows/2d9de7e8-6986-49c0-8274-e41b91180e2b/jobs/8433?invite=true#step-108-20270_139 ).

I guess we're on a good way but our release is not finished.

**Comment by kkimurak** [comment]
I've had some very rare issues with `puma` dying frequently, but restarting the container almost always solved it.  
Also could not reproduced with official postgres:17 image and my kkimurak/sameersbn-postgresql:17 image (after deleting all docker volumes) so I believe postgresql is not a cause.  
If you have any tips on how to reproduce the problem or the actual error message, please let me know. I'll look into it.

**Comment by sachilles** [comment]
I did build the docker image and run the docker compose mircoservices. In the logs I found

```
gitlab-1      | 2025-05-18 03:15:56,501 WARN exited: gitaly (exit status 1; not expected)
gitlab-1      | 2025-05-18 03:15:58,792 INFO spawned: 'gitaly' with pid 874
gitlab-1      | 2025-05-18 03:15:59,166 WARN exited: gitaly (exit status 1; not expected)
gitlab-1      | 2025-05-18 03:16:02,246 INFO spawned: 'gitaly' with pid 897
gitlab-1      | 2025-05-18 03:16:02,581 WARN exited: gitaly (exit status 1; not expected)
gitlab-1      | 2025-05-18 03:16:03,582 INFO gave up: gitaly entered FATAL state, too many start retries too quickly
```

which shows up in the admin interface as 

grafik

**Comment by sachilles** [comment]
After cleaning up the docker volumes the issue wasn't present anymore.

**Comment by sachilles** [comment]
@kkimurak If there are no further complaints, I will create the release.

**Comment by kkimurak** [comment]
@sachilles 

Glad it works.  

Documentations need to be updated to note gitlab 18.x or later requires postgresql 16.x. It appears that you have already updated release-note.sh, so I think you only need to fix two places in README.md. 

Ref (gitlab 17.x and postgresql 14.x : Pull request #3001 commit 6f43b19)

**Comment by sachilles** [comment]
@kkimurak and @th-2021 Thanks for supporting this project.

**Comment by th-2021** [comment]
I ran into the gitaly issue as well and it turned out to be a filesystem error. The "git" user from the new image is now 1001:1001 , while it was 1000:1000 in previous images. So I had to chown the volumes.

**Comment by th-2021** [comment]
1000:1000 is now "ubuntu", which is not needed in our case. So we could remove it. What do you think?
The purpose of "ubuntu" is to have a login user in the cloud image.

**Comment by sachilles** [comment]
@th-2021 If we don't need the user 'ubuntu' then we should remove the account.

**Comment by th-2021** [comment]
in assets/runtime/functionsL2263: max-old-space-size should be 8192 (to match install.sh)
    exec_as_git bundle exec rake gitlab:assets:compile NODE_OPTIONS="--max-old-space-size=8192" >/dev/null 2>&1

**Comment by sachilles** [comment]
**Note:** GitLab came up without any errors. However pushing code to a repository isn't possible and results in `fatal: protocol error: bad line length character`. Do you have any ideas?

**Comment by sachilles** [comment]
In Addition I found that the GitLab runners don't pick up any jobs.

Furthermore, running all checks shows that sidekiq causes some issues.

**Comment by kkimurak** [comment]
@sachilles `ssh` to gitlab service shows some message like `This account is currently not available`. If the `ssh` command echos some string to console, git action with ssh fails with a message like `error: bad line length character: This` ('This' is the first word in the message above).

```
$ ssh git@localhost -p 10022 -i ~/.ssh/id_rsa_test
PTY allocation request failed on channel 0
This account is currently not available.
Connection to localhost closed.
```

It seems the user `git` does not allowed to login as the login shell is `/usr/bin/nologin`. Could you try removing `--disabled-login` option from adduser to check if works.

In running 17.11.2  
```
$ docker exec -it gitlab cat /etc/passwd | grep git
git:x:1000:1000:GitLab,,,:/home/git:/bin/bash
```

Locally built 18.0.0
```
$ docker exec -it gitlab-pr3107 cat /etc/passwd | grep git
git:x:1000:1000:GitLab,,,:/home/git:/usr/sbin/nologin
```

sachilles:
@kkimurak Thanks for pointing into the right direction. Manually correcting the entry in `/etc/passwd` helps. I'll prepare a commit and check this out.

sachilles:
@kkimurak There seems to be some changes since in ubuntu 24.04 the option `--disabled-login` will additionally set the shell to `/usr/sbin/nologin`.

I guess that the option `--disabled-password` should be used (in ubtuntu 24.04). The manpage for `adduser` states: [...] Do not run passwd(1) to set a password.  In most situations, logins are still possible though (for example using SSH keys or through PAM) for reasons that are beyond adduser's scope.

In ubuntu 22.04 the option `--disabled-login` there is no statement about modifications in the users shell. However, in both distributions the option `--disabled-passwd` is present and (based on the manpage contents) means the same.

kkimurak:
Oh I see..

https://manpages.ubuntu.com/manpages/jammy/man8/adduser.8.html
https://manpages.ubuntu.com/manpages/noble/man8/adduser.8.html

The chagne seems to be imported from debian: https://salsa.debian.org/debian/adduser/-/merge_requests/79

sachilles:
Finally, the last build of this PR seems to work. I have tested the following functions:

- Login
- Pull and push to Git repositories
- Pull and push of Docker images to the registry

From my point of view, there is nothing to be said against the corresponding release or mixing of this PR.

kkimurak:
Great. If there is anything to be worried about, it would be around backups (because I have not tested yet).

sachilles:
Taking a backup was tested by myself. But restoring from backup not.

sachilles:
I think we should release this version and if some problems are reported (hopefully) we'll manage to fix them.

kkimurak:
Maybe not. Let's pray:)

> Shall we install the replacement for runit-systemd, or not?

PR #3116, opened 2025-05-29
Remove version from compose files

aminvakil:
This will prevent this message at every `docker compose SOMETHING`.
the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion

### DIFF
--- a/README.md
+++ b/README.md
@@ -2807,8 +2807,6 @@ To do that you will need to set the environment variable `GITLAB_MONITORING_IP_W
 You can also set your `docker-compose.yml` [healthcheck](https://docs.docker.com/compose/compose-file/compose-file-v2/#healthcheck) configuration to make periodic checks:
 
 ```yml
-version: '2.3'
-
 services:
   gitlab:
     image: sameersbn/gitlab:18.0.1
--- a/contrib/docker-swarm/docker-compose.yml
+++ b/contrib/docker-swarm/docker-compose.yml
@@ -1,5 +1,3 @@
-version: '3.4'
-
 services:
   redis:
     restart: always
--- a/docker-compose.swarm.yml
+++ b/docker-compose.swarm.yml
@@ -1,4 +1,3 @@
-version: '3.4'
 services:
   redis:
     image: redis:7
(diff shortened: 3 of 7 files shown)


kkimurak:
Removing `version` top-level element can be applied uniformly to other files as well.

https://github.com/sameersbn/docker-gitlab/blob/37a0b9506ffefaf7ec74961a6ed936d2ec525174/docker-compose.swarm.yml#L1

https://github.com/sameersbn/docker-gitlab/blob/37a0b9506ffefaf7ec74961a6ed936d2ec525174/contrib/docker-swarm/docker-compose.yml#L1-L2

https://github.com/sameersbn/docker-gitlab/blob/37a0b9506ffefaf7ec74961a6ed936d2ec525174/docs/docker-compose-registry.yml#L1-L2

https://github.com/sameersbn/docker-gitlab/blob/37a0b9506ffefaf7ec74961a6ed936d2ec525174/docs/docker-compose-keycloak.yml#L1-L2

Also in docs.

https://github.com/sameersbn/docker-gitlab/blob/37a0b9506ffefaf7ec74961a6ed936d2ec525174/README.md?plain=1#L2810-L2811

https://github.com/sameersbn/docker-gitlab/blob/37a0b9506ffefaf7ec74961a6ed936d2ec525174/docs/s3_compatible_storage.md?plain=1#L68-L69

----

Not sure if we should apply the document-start rule (start with three hypens) to compose.yml as far at least official docker docs does not describes to follow the rule (I also have seen [some](https://github.com/WeblateOrg/docker-compose/blob/b4339acae26029b2e2de512c451e559e3c49f185/.yamllint.yml#L13) [repositories](https://github.com/AxisCommunications/docker-compose-acap/blob/28adaeb6272b2d2922d43e65f4ce92a5e985c2c2/.yamllint.yml#L8) setting this rule to be "disabled" in yamllint config).  

The docker compose (v2.36.2) installed in my environment seems to work fine with or without the three hyphens at the beginning of the file, so I think it doesn't matter which one we choose.

sachilles:
@aminvakil Thanks for contributing to this project.

PR #3121, opened 2025-06-17
Allow setting ActiveRecord encryption secrets

### LINKED ISSUE #3112: Bug: Active Record secrets impossible to persist, making certain GitLab features like Dependency Proxy unusable
GitLab added a few new secrets to store operational secrets in the database via Active Records (see [Operational Secrets](https://docs.gitlab.com/development/application_secrets/#secret-entries)). These three keys are the relevant pieces in the encryption:

* `active_record_encryption_primary_key`
* `active_record_encryption_deterministic_key`
* `active_record_encryption_key_derivation_salt`

The first two are lists, as I understand, with the third being a simple value. When GitLab starts and detects that they are unset, it generates them like so (in `/home/git/gitlab/config/secrets.yml` - don't worry, those are not my actual keys):

```yaml
  active_record_encryption_primary_key:
  - 1OmFswg5dUL5Cgl6i3CTVm3bUFXMDC42
  active_record_encryption_deterministic_key:
  - 5e7DiXxhy3g73Nd8oo4YuZwbLIW7XnZ3
  active_record_encryption_key_derivation_salt: ZAATKYJSHgr0DgrfX5d27LEeBs6gOGbH
```

Unlike other entries in that file such as `secret_key_base`, `otp_key_base` or `encrypted_settings_key_base`, the Active Record keys currently can not be set at all by this repository, not even through the normal process of providing them as ENV vars. I've looked into how the file is populated/updated by this repo, and from my understanding, no avenue to provide your own secrets is currently possible, because the file is generated from `/etc/docker-gitlab/runtime/config/gitlabhq/secrets.yml` which as you can see in https://github.com/sameersbn/docker-gitlab/blob/master/assets/runtime/config/gitlabhq/secrets.yml simply has no entries for those values (and when I tried to hack entries in there, they were not populated, so further work might also be needed).

I suggest that the same process that lets us set other secret variables through ENV vars such as `GITLAB_SECRETS_SECRET_KEY_BASE` and friends is expanded to cover these three new variables, as without such a change, features such as the Docker Dependency proxy are currently unusable - the feature gets corrupted every time the container restarts. I will file another issue with details on how to recover from that happening.

The only workaround I can think of to use such features at all with this container right now is to start the container, display the values of what GitLab generates (once it has been running for a minute) like so:

`docker exec -it gitlab cat /home/git/gitlab/config/secrets.yml`

Then write those copies to a file that is then mounted in place as a volume via `/your/path/secrets.yml:/home/git/gitlab/config/secrets.yml` with whatever permissions and settings are appropriate in your setup. WARNING: this will break various features of this repository by hard coding what is usually set by the above mentioned variables!

kkimurak:
Fix #3112

This PR allow setting ActiveRecord encryption secrets via environment variables `GITLAB_SECRETS_ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY`, `GITLAB_SECRETS_ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY` and `GITLAB_SECRETS_ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT`.

First implementation by me, documentation correction and test by @Gaibhne .


sachilles:
@kkimurak and @Gaibhne Thank your for your contribution.

PR #3122, opened 2025-06-17
Stop recompiling assets to enable relative url

### LINKED ISSUE #2975: Relative URL Does Not Work If There Is No Access to NPM Registry
When I use GITLAB_RELATIVE_URL_ROOT It reinstalls all the Node modules ([Link to Code](https://github.com/sameersbn/docker-gitlab/blob/362f6a06bc4afbbbe0b0477c5060e544ff9d48a5/assets/runtime/functions#L2257)).

However, when running this in an airgapped environment without access to the NPM registry, it gets stuck because it cannot find the modules to install.

**Steps to Reproduce:**
Set up an airgapped environment without NPM registry access.
Use the GITLAB_RELATIVE_URL_ROOT at the specified location in the code.

**Expected Behavior**
The Node modules should not attempt to reinstall, or there should be an alternative method to handle this in an airgapped environment.

**Actual Behavior:**
The process gets stuck and cannot find the modules to install.

Is there a workaround or an alternative approach to handle this situation in an airgapped environment?

link to another issue where i found it - https://github.com/sameersbn/docker-gitlab/issues/2964#issuecomment-2247506803

kkimurak:
To enable relative url, we execute rake task [`gitlab:assets:compile`](https://github.com/sameersbn/docker-gitlab/blob/0d39d55df305d17b1bd0d74b654cb41aaa8db51f/assets/runtime/functions#L2254-L2264). This step have been removed in official documentation in v10.0.0 (first committed to gitlab-foss, then ported to gitlab(-ee))
- EE: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/1139/
- CE: https://gitlab.com/gitlab-org/gitlab-foss/-/merge_requests/8831

This PR simply remove this runtime process.  
This change will reduce startup time when the relative URL feature is enabled (when compiling assets, for example when performing a version upgrade or changing relative URLs). In my environment, the startup time, which previously took about 20 minutes, has now been reduced to about 1 minute.

Also, we no longer need to do following things
- persist node_modules, recompiled assets and caches
- apply patches to gitlab itself to avoid removing (symlink to) node_modules/

As a result, this PR allow us to enable relative URL without internet connection.  
fix #2975 .


sachilles (approved):
LGTM

sachilles:
@kkimurak Als always, thank you for your contributions. 🌹

PR #3127, opened 2025-06-19
docs: Fix required PostgreSQL version for GitLab 18.0.0 (was 18.0.2)

mixxtor:
Updated the documentation to correctly reflect that GitLab 18.0.0 requires PostgreSQL 16.x, not 18.0.2 as previously stated.


sachilles:
@mixxtor Thanks for the fix.

PR #3133, opened 2025-07-03
Use ed25519 in expose-gitlab-ssh-port.sh

aminvakil:
It does not break systems configured with `id_rsa` prior to this PR, it will get updated to `id_ed25519`, but previous `id_rsa` key remains there which I suppose is not a problem.

### DIFF
--- a/contrib/expose-gitlab-ssh-port.sh
+++ b/contrib/expose-gitlab-ssh-port.sh
@@ -10,15 +10,15 @@ if ! id -u git >> /dev/null 2>&1; then
 fi
 su git -c "mkdir -p /home/git/.ssh/"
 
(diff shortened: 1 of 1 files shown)


sachilles:
@aminvakil Is there any specific reason for changing the algorithm for generating the used ssh keys?

However, in favor of fast updates of the provided GitLab versions via this project I skipped (temporarily) your PR. Could you be so kind an rebase your branch onto the branch `master`?

aminvakil:
> @aminvakil Is there any specific reason for changing the algorithm for generating the used ssh keys?
> 

Not a particular reason besides very insignificant increase in speed in ed25519 and that it is mathematically harder to break than RSA.

It's been long supported (since 2014 in openssh) and it's best practice to use it whenever possible instead of RSA.

> However, in favor of fast updates of the provided GitLab versions via this project I skipped (temporarily) your PR. Could you be so kind an rebase your branch onto the branch `master`?

Done

Issue #2860, opened 2023-11-27
Improve image build speed (Discussion,Ideas)


gjrtimmer:
This is a discussion for a concept I'm working on; it is also preliminary work on hopefully support for arm64.
I'm working on a concept to improve the build speed of the image.

The concept I'm using within my work and in my private repositories has a base image automatically updated with a CI cron job at least once a week. This image is written to the registry with the tag `base.` The next stage of the build is simply using this image tag.

Furthermore, I want several time-consuming build stages of this image to be extracted to speed up the building process.
For example, ruby 3.0.6 is used throughout several versions. However, building this every time is time-consuming. 
Therefor, I'm looking towards having this specific ruby bing built by a CI process, which then writes it to the repository as an asset as a *.deb file stored in the repository under git-lfs. The install stage has to checkout the repository, perform git-lfs checkout, start the build context, and then only install the *.deb package. This can also be done with the other compilation parts.

This is a rough draft; I'm looking forward to input and comments.

Please let me know what you think.

Disclaimer: Due to my profession, I'm a hefty GitLab user; I can quickly build all of this in GitLab; however, building it within GitHub actions will be a challenge and probably require help and input from others.

gjrtimmer:
@sameersbn @solidnerd question is the reason for compiling ruby the --enable-shared parameter? Is this not default, second question can we not use the latest ruby 3.2.2?

Im looking to speed up the build process so I'm looking why the image compiles ruby itself

kkimurak:
I authored the change (Install ruby from source - #2429) to fill gitlab operation requirement on 14.3.0 (ruby 2.7.4).  
There is no clear reason why I chose the source installation. At the time, I was just in a hurry. Around GitLab 13, there was a significant delay in keeping up with GitLab releases, and I was afraid that the project would stall. This remains true even today.

Early releases of sameersbn/gitlab just put binaries (release tarballs, rubygem caches and so on) but it have been revised by sameersbn - see https://github.com/sameersbn/docker-gitlab/pull/30#issuecomment-40192744

Through a few years [brightbox/ruby-ng](https://launchpad.net/~brightbox/+archive/ubuntu/ruby-ng) was used as ruby package installation source but it has stopped updating,

I once tried porting the ruby installation process to a Dockerfile in the hopes that the cache would be utilized. But for local builds, the speedup was only a few minutes.

[...]

PR #3138, opened 2025-07-10
Disable compiling sourcemaps to reduce build time

kkimurak:
This PR just adds image ENV `NO_SOURCEMAPS=true`.  

As I reported in #2860 (https://github.com/sameersbn/docker-gitlab/issues/2860#issuecomment-3012732268), we can set webpack configuration [`devtool`](https://webpack.js.org/configuration/devtool/) to `false` by setting environment variable [`NO_SOURCEMAP=true`](https://gitlab.com/gitlab-org/gitlab/-/blob/v18.1.2-ee/config/webpack.config.js#L952). This will significantly reduce the time taken by the rake task `gitlab:assets:compile`, which took up a large portion of the image build (total build time have been improved from 30min to 20min on my env).  

It also reduce image size : v18.1.2, 4.31GB -> 3.86GB.

In webpack documentation, it is recommended to set `devtool` to be disabled on production with maximum performance. 

> | devtool | performance | production | quality | comment
> | -- | -- | -- | -- | --
> | (none) | build: fastest<br>rebuild: fastest | yes | bundle | Recommended choice for production builds with maximum performance.

Official gitlab image `gitlab/gitlab-ce` and `gitlab/gitlab-ee` also does not ship with sourcemaps (see https://github.com/sameersbn/docker-gitlab/issues/2860#issuecomment-3056325680).  

Since I have not yet been able to verify how much each factor, such as CPU, memory, or disk I/O, affects the speed improvement, I would like to check the build time of this pull request to see how much the speed has improved in CI.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,8 @@ ENV GITLAB_VERSION=${VERSION} \
     GITLAB_LOG_DIR="/var/log/gitlab" \
     GITLAB_CACHE_DIR="/etc/docker-gitlab" \
     RAILS_ENV=production \
-    NODE_ENV=production
+    NODE_ENV=production \
+    NO_SOURCEMAPS=true
 
 ENV GITLAB_INSTALL_DIR="${GITLAB_HOME}/gitlab" \
     GITLAB_SHELL_INSTALL_DIR="${GITLAB_HOME}/gitlab-shell" \


kkimurak:
Build job [8731](https://app.circleci.com/pipelines/github/sameersbn/docker-gitlab/2006/workflows/ffe4006c-f096-4996-aa2e-1a91a28ca29e/jobs/8731) completed in 26min 54sec. Compared to [8696](https://app.circleci.com/pipelines/github/sameersbn/docker-gitlab/1999/workflows/9e21eff4-07cc-4adc-a786-bfaf0b93f7ad/jobs/8696) (18.1.2 release pull request #3137) tooks 39m 44s, reduced about 13 minutes.

kkimurak:
Well, at this point I'm not sure if this pull request is the right direction.

It definitely reduces build times and image size, by not including sourcemaps.

I thought I understood how much of an impact it would have, but after reviewing the upstream discussion I'm not sure again.

----

I have noticed that official images `gitlab/gitlab-ce` and `gitlab/gitlab-ee` does not ship with **uncompressed** .map files but with **compressed** map files (*.gz). Uncompressed files are just [removed](https://gitlab.com/gitlab-org/omnibus-gitlab/-/blob/ea0ad50a32a337bbd0e135c3ba10eb5666a709a1/config/software/gitlab-rails.rb#L217-218) on image built, since 12.10.0 (introduced by [gitlab.com/gitlab-org/omnibus-gitlab!3592](https://gitlab.com/gitlab-org/omnibus-gitlab/-/merge_requests/3592)).  
CNG (Cloud Native container images per component of GitLab) takes similar approach. See [gitlab.com/gitlab-org/build/CNG!613](https://gitlab.com/gitlab-org/build/CNG/-/merge_requests/613).  

This means that assets need to be compiled once, which takes so long that I initially thought I would disable it, thus making this pull request pointless.

I found evidence that `NO_SOURCEMAPS` was set in image builds for low memory enviornment such as Raspberry Pi around version v11.0.2, but this setting is no longer present - I was unable to determine whether the setting has been removed or is being managed in a different way. See [gitlab.com/gitlab-org/omnibus-gitlab!2606](https://gitlab.com/gitlab-org/omnibus-gitlab/-/merge_requests/2606)

GitLab.com seems to be providing source maps (via CDN) to avoid confusing users seeing a lot of 404 on console. See discussions for example: [gitlab.com/gitoab-org/gitlab#212240](https://gitlab.com/gitlab-org/gitlab/-/issues/212240).

sachilles:
@kkimurak Thanks for rebasing and investigating. And of course for your valuable work! 🌻

PR #3186, opened 2025-11-13
Maintenance on PostgreSQL clients

sachilles:
Since PostgreSQL v18 was already published we need to provide a suitable postgresql client as well.

However, following the requirements for GitLab CE and in particular for PostgreSQL (see <https://docs.gitlab.com/18.5/install/requirements/#postgresql>) we should remove all clients for PostgresSQL versions smaller than 16.


kkimurak:
I think it's ok to remove old PostgreSQL clients. In addition to release notes, how about displaying a warning (or terminating with an error) if the minimum and maximum version requirements are not met?
The process for obtaining the server version has already been implemented, so changing it to the following should generally work (see my rough implementation: [check-potgresql-server-version-and-wan](https://github.com/kkimurak/docker-gitlab/tree/check-postgresql-server-version-and-warn) branch based on this PR, especially commit  7199c085ea021b292f069549985e7f7356d8c850)

By the way, according to the link, Maximum PostgreSQL version for GitLab 18.x is 17.x ([tested against GitLab 17.10 and later](https://gitlab.com/gitlab-org/gitlab/-/issues/521159))

sachilles:
@kkimurak Since GitLab version 19 (see #3250) requires at least PostgreSQL version 17, I would very much like to consider this PR. 

Since I am not yet satisfied with the current state of the PR, I would modify it as follows:

- If the existing PostgreSQL server version is lower than the version required by GitLab, the container **should not start**, and a message indicating the necessary PostgreSQL update should appear.
- If the existing PostgreSQL server version is higher than the one required by GitLab, only a warning should be displayed; the container **should not be prevented** from starting.

What do you think?

Dockerfile
```diff
@@ -16,7 +16,10 @@ ENV GITLAB_VERSION=${VERSION} \
     GITLAB_CACHE_DIR="/etc/docker-gitlab" \
     RAILS_ENV=production \
     NODE_ENV=production \
-    NO_SOURCEMAPS=true
+    NO_SOURCEMAPS=true \
+    # v18.5.2 : minimum = 16.5, maximum = 17.x (currently 17.6, is 170006)
```
kkimurak: Current postgresql 17.x release is 17.10 (170010) so my comment in code is stale a bit (`(currently 17.6, is 170006)`). If you find future maintenance troublesome, you might consider deleting the line.

```sh
$ docker run --rm -d --name psql17_test kkimurak/sameersbn-postgresql:17-20260521
b41212df4b1843b50091f6d452665847e3decc4f89dd2e9cff87af18077ab753

$ docker exec -it -u postgres psql17_test psql -h localhost -c 'SHOW server_version_num;'
 server_version_num 
--------------------
 170010
(1 row)
```


```suggestion
    # v18.x : minimum = 16.5, maximum = 17.x (currently 17.10, is 170010)
```

assets/runtime/functions
```diff
@@ -190,13 +190,38 @@ gitlab_generate_postgresqlrc() {
 
   echo "- Detected server version: ${DB_SERVER_VERSION}"
 
+  # remove leading zero (prior to 10.x it was like "090605" so that cannot treated as number)
+  DB_SERVER_VERSION="${DB_SERVER_VERSION##+(0)}"
```
kkimurak: Tested on [wandbox](https://wandbox.org/permlink/BFQAiVBtbXWLSEkK) and noticed it doesn't work as expected  
(I doubt anyone would bother using it in combination with PostgreSQL 9.x, though...)

```suggestion
  DB_SERVER_VERSION="${DB_SERVER_VERSION#0}"
```

kkimurak (changes requested):
@sachilles Thank you for your work and sorry for delay, generally looks good to me.  
I found some mistake in my code. I don't think these will be a problem, but I'll correct them just in case.

assets/runtime/functions
sachilles: Thanks for pointing this out.

Dockerfile
sachilles: Took your comment and since rebased onto branch `upgrade-to-19.0.0` adapted your suggestion for GitLab v19.x.

sachilles:
@kkimurak Thanks for your additional review!

kkimurak (approved):
My home wired network connection is unavailable until the end of the month, and I only have a slow, data-limited connection, so I can't even fetch GitLab codebase and therefore can't build and test the image.  
Anyway, from what I've quickly checked, it looks good to me.

sachilles:
@kkimurak Thanks anyway!

PR #3213, opened 2026-02-07
Fix pages when access control is true

uda:
When `GITLAB_PAGES_ACCESS_CONTROL` is set to `true` the variable `GITLAB_RELATIVE_URL_ROOT` added recently is not being replaced causing pages to fail starting


emotroshylov:
Hi @sachilles 
Could you please take a look at this PR when you have some time?
This bug is still fully reproducible and completely breaks GitLab Pages on recent versions like v19.0.1 whenever GITLAB_PAGES_ACCESS_CONTROL is enabled. 
This PR contains a solid fix that solves the problem. It would be awesome to get this reviewed and merged. 
Thank you!

sachilles:
@emotroshylov Thanks for the ping. I'm creating backports for all the other active versions.

sachilles:
@uda Thank you for your contribution, and I apologize for the long wait.

PR #3236, opened 2026-04-16
Cronjob vars

th-2021:
I added variables for two more cronjobs as they contain an hour setting (which doesn't work for me). I synced the cronjob section also with upstream


sachilles:
LGTM

PR #3295, opened 2026-08-20
Enable HSTS for GitLab Pages

### LINKED ISSUE #2710: The NGINX_HSTS_ENABLED setting is not valid for Pages
Pages could not turn on HTTP Strict Transport Security for NGINX.
Modifying NGINX_HSTS_ENABLED doesn't seem to work for Pages.

loglapa:
## What changed

- Add the existing HSTS header template to both GitLab Pages nginx configurations.
- Render the configured `NGINX_HSTS_MAXAGE` when Pages HTTPS is enabled.
- Remove the header when Pages HTTPS or `NGINX_HSTS_ENABLED` is disabled.
- Clarify in the README that the existing HSTS options also apply to GitLab Pages.

## Why

The main GitLab nginx virtual host honors `NGINX_HSTS_ENABLED` and `NGINX_HSTS_MAXAGE`, but the Pages virtual hosts do not contain or configure the header. Changing those options therefore has no effect on Pages responses.

The non-TLS Pages template also needs the placeholder for deployments where TLS terminates at a load balancer while `GITLAB_PAGES_HTTPS=true`, matching the existing main GitLab virtual host behavior.

## Impact

GitLab Pages responses now use the same HSTS settings as the main GitLab virtual host, without adding new environment variables or changing the defaults.

## Validation

- `bash -n assets/runtime/functions`
- `git diff --check`
- Rendered and checked four cases: native Pages HTTPS, HSTS disabled, plain HTTP, and HTTPS terminated by a proxy

Fixes #2710

### DIFF
--- a/README.md
+++ b/README.md
@@ -1868,11 +1868,11 @@ Sets the bucket size for the server names hash tables. This is needed when you h
 
 ##### `NGINX_HSTS_ENABLED`
 
-Advanced configuration option for turning off the HSTS configuration. Applicable only when SSL is in use. Defaults to `true`. See [#138](https://github.com/sameersbn/docker-gitlab/issues/138) for use case scenario.
+Advanced configuration option for turning off HSTS for the GitLab and GitLab Pages nginx virtual hosts. Applicable only when SSL is in use. Defaults to `true`. See [#138](https://github.com/sameersbn/docker-gitlab/issues/138) for use case scenario.
 
 ##### `NGINX_HSTS_MAXAGE`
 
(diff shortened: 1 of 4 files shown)


sachilles:
@loglapa I'd like to review your PR. However, could you be so kind to revert this branch to 9e7e35c7ddc9251f88a36ca0604c0ffdcd45ee73 and rebase this onto the current master?

Thank you in advance.

loglapa:
Done — restored the branch to the original change and rebased it onto the current `master`. The new signed commit is `2df80e0`; CircleCI has restarted.

PR #3302, opened 2026-09-11
Upgrade GitLab CE to 19.3.2

sachilles:
GitLab Critical Patch Release: 19.3.2 (for more details see <https://docs.gitlab.com/releases/patches/patch-release-gitlab-19-3-2-released/>)


sachilles:
@sameersbn We need your help. In the final step of the pipeline—that is, when generating the release notes—all pipelines are currently stopping with the following error message:

> error: github returned 401 Unauthorized
> Exited with code exit status 1

Please check, if the GITHUB_TOKEN is active and not expired.

sachilles:
@sameersbn We **urgently** need your help. Without your help, we won't be able to create any more releases!

kkimurak:
FYI: I tested it locally against my own `kkimurak/docker-gitlab` repository following release section in `.circleci/config.yml` and was able to create a release without any issues, so it doesn't seem to be a problem caused by using the staled `meterup/github-release`. In other words, there doesn't appear to be anything I can do about it.

Personally, I think that in a worst-case scenario, we could do without release notes (as long as the Docker image is published). If necessary, we could - though it might be a bit messy - create a file like `releases/v19.x/v19.3.1.md` containing the details from the release note script?

sachilles:
@kkimurak I'll create the release notes by hand via the web interface of Github.

kkimurak:
@sachilles It is a shame that we have moved a step away from the automated repository. Still, I suppose it is better than having updates stall..  
Anyway, thank you for handling this; I feel bad about the increased burden on you.

